PR #24, opened 2023-07-17
archives tool prototypes

francoismg:
Tools prototypes for archives browsing with esap and pyvo


tools/archives/esap_integration/astronomical_archives.py
```diff
@@ -0,0 +1,476 @@
...
+        'dec': '',
+        'fov': '',
+        'collection': '',
+        'page': '',
+        'archive_uri': 'astron_vo'
+    }
+
+    def __init__(self):
+        super().__init__()
+        self._base_url = "https://sdc-dev.astron.nl/esap-api/query/query/"
```
volodymyrss, line 66: we really do not want to query ESAP. I talking about ESAP as an example of how a platform like galaxy can query IVOA archives. You may note that eventual archivess tool in galaxy looks a bit like ESAP IVOA query in functinality.

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,55 @@
...
+        <option value="euv">EUV</option>
+        <option value="x-ray">X-ray</option>
+        <option value="gamma-ray">Gamma-ray</option>
+      </param>
+      <param name="query" type="text" label="Query" value="SELECT TOP 100 * from ivoa.obscore" />
+    </inputs>
+    <outputs>
+        <data name="output" format="fits" />
+        <data name="output_url" format="tabular" />
+    </outputs>
```
volodymyrss, line 49: if you add tests, it will also serve as an example

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,55 @@
...
+    <help><![CDATA[
+At the moment this tool allows the following simple searches:
+
```
volodymyrss, line 52: some small readme help would be great even at this stage

tools/archives/pyvo_integration/astronomical_archives.py
```diff
@@ -0,0 +1,300 @@
...
+    tabular_output = sys.argv[8]
+
+    file_url = []
+
+    rsp = RegistrySearchParameters(keyword=keyword, waveband=waveband, service_type='TAP')
+
+    archive_list = Registry.search_registries(rsp, 1)
+
+    if archive_list:
+        archive_list[0].initialize()
```
volodymyrss, line 286: so you will just query fro the first archive? It's not what users wil want.

Did you not have a version were it was possible to select archives as input?

francoismg:
> Thanks, very good to see the code! We tried to look into it, but something did not work. Let's see along the PR.
> 
> One thing which would help, by the way, is tests. Could you add tests like galaxy tools suggest? See my tools in the same repo for example.
> 
> Also, some help would be useful just as an explanation.
> 
> I'll add comments in the code.

Was there a problem in installing the tool or when running it? What was the problem?

I will look into adding tests no problem, I had thought about it but since the main goal is to download files the test should be able to compare a downloaded file with a file that would be provided for testing and since file can change depending on what archives are added to the registry that cause some problem but I will look into it.

Will also add some help part too of course

tools/archives/esap_integration/astronomical_archives.py
francoismg: I thought the goal was to integrate esap since it's  also a wp5 project. Since anyone can put some esap gateway online and add some specific archive I thought it would be good since it would make it possible to integrate more esap gateways with different archives that might not be accessible through ivoa (like cta if I'm not mistaken) also like with the astron archive it enables the archive to put more specific parameters (like collections) that I wasn't able to find with pyvo.


If you don't want esap we can use only the pyvo integration but what I had in mind was that the two could be merged to either work as fallback if no esap gateway are available or to let the user choose if he wants to query an esap gateway or use pyvo directly.

volodymyrss:
> > Thanks, very good to see the code! We tried to look into it, but something did not work. Let's see along the PR.
> > One thing which would help, by the way, is tests. Could you add tests like galaxy tools suggest? See my tools in the same repo for example.
> > Also, some help would be useful just as an explanation.
> > I'll add comments in the code.
> 
> Was there a problem in installing the tool or when running it? What was the problem?

It runs but the inputs we put the result is empty. It means that we put not what you do, but it also means that it's not intuitive for us and we need to eventually end up in a situation when it is. It's ok if not for now.

> I will look into adding tests no problem, I had thought about it but since the main goal is to download files the test should be able to compare a downloaded file with a file that would be provided for testing and since file can change depending on what archives are added to the registry that cause some problem but I will look into it.

Archive should be part of the input. Ideally, it should be parameteric input somehow, showing archives to select depending on the input keywords. You showed something like this, right?

I am thinking that if it is not possible to query list of archives dynamically from the tool, then we have two options:
1. keep static list of archives in the tool. i.e. query it at the time of the tool delivery and store in the tool dir. we can update it with new tool versions. List of archives changes, but not too fast.
2. separate tool in two: fetching list of arhives; fetching files/urls from an archive. it could be the same tool also, with two modes of operation.

> Will also add some help part too of course

thanks!

tools/archives/pyvo_integration/astronomical_archives.py
francoismg: The selection of specific archives rely on the possibility to dynamically create the tool interface which I wasn't able to do (I think there be might a way (as the upload tool loads a list of datatypes from calling some galaxy registry file) but I wasn't able to figure it out and also i'm pretty sure it would require adding some code to core galaxy which can't be done if it's a simple tool in the toolshed)

Another possibility would be to download every archive tap service url into a tool data file and populate a list from there which I think could work but it means the list should be regularly updated to account for changes in archives

tools/archives/esap_integration/astronomical_archives.py
volodymyrss: > I thought the goal was to integrate esap since it's also a wp5 project. Since anyone can put some esap gateway online and add some specific archive I thought it would be good since it would make it possible to integrate more esap gateways with different archives that might not be accessible through ivoa (like cta if I'm not mistaken) also like with the astron archive it enables the archive to put more specific parameters (like collections) that I wasn't able to find with pyvo.

There is not really a guarantee that there will be many ESAP interfaces. 
The baseline interface which is common, and will be adopted in some way across more current project of ours, is the IVOA interface (interfaces).

If pyvo does not support some necessary part of IVOA protocols, you can try to implement it yourself in python following specification. Or even propose a change to pyvo.
If esap supports something which is not supported by IVOA, then we can not afford to rely on such a feature since there is no guarantee that this feature will remain supported. So missing functionality should be added in the client tool.

> 
> If you don't want esap we can use only the pyvo integration but what I had in mind was that the two could be merged to either work as fallback if no esap gateway are available or to let the user choose if he wants to query an esap gateway or use pyvo directly.

Relying on ESAP will have peculiar and in most cases unfortunate implications of all kinds, most of all with sustainability. Since you already implemented something which relies on ESAP, I suggest you move it to a separate tool and not put much effort on it anymore. We probably could use this experience, but it's something very different.

Meanwhile, we really need a tool which works on IVOA protocols directly, like pyvo does. Note that there are also  IVOA services not in this registry you see. You can use pyvo when suitable, but most of all please rely on protocols.

tools/archives/pyvo_integration/astronomical_archives.py
volodymyrss: The second possibility is what I also suggested as a first implementation, let's do this, and then ask our Galaxy colleagues for advice.

francoismg:
> > > Thanks, very good to see the code! We tried to look into it, but something did not work. Let's see along the PR.
> > > One thing which would help, by the way, is tests. Could you add tests like galaxy tools suggest? See my tools in the same repo for example.
> > > Also, some help would be useful just as an explanation.
> > > I'll add comments in the code.
> > 
> > 
> > Was there a problem in installing the tool or when running it? What was the problem?
> 
> It runs but the inputs we put the result is empty. It means that we put not what you do, but it also means that it's not intuitive for us and we need to eventually end up in a situation when it is. It's ok if not for now.
> 
> > I will look into adding tests no problem, I had thought about it but since the main goal is to download files the test should be able to compare a downloaded file with a file that would be provided for testing and since file can change depending on what archives are added to the registry that cause some problem but I will look into it.
> 
> Archive should be part of the input. Ideally, it should be parameteric input somehow, showing archives to select depending on the input keywords. You showed something like this, right?
> 
> I am thinking that if it is not possible to query list of archives dynamically from the tool, then we have two options:
> 
>     1. keep static list of archives in the tool. i.e. query it at the time of the tool delivery and store in the tool dir. we can update it with new tool versions. List of archives changes, but not too fast.
> 
>     2. separate tool in two: fetching list of arhives; fetching files/urls from an archive. it could be the same tool also, with two modes of operation.
> 
> 
> > Will also add some help part too of course
> 
> thanks!

For the inputs part did you remember what input did you try? I can try to make it more intuitive if I know how you were trying to use it.

For the specific archive selection I responded in a previous comment where I basically suggest the same thing you do (using a pre genereated list).

Separate the tool in two could be done too but then I guess it would be really useful to have some plugin that let you interact with the archives files to see every archives description, tap schema and more with the possibility to edit the file to mark which archive you want to query

tools/archives/esap_integration/astronomical_archives.py
francoismg: Ok I will switch to the pyvo version and try to depend as less as possible on pyvo and more on directly interacting with the protocol. I didn't thought you wanted some from scratch implementation since it represents a good part of the pyvo codebase but I think it could be done. I tried that way at first so I have some ideas on how it could work, will work on it this weekend so I can get back to you on that

volodymyrss: Well whatever pyvo provides is good to use. If it turns out that it misses some important features we can propose to add them actually, it will be much appreciated by the community.

Certainly there is no need to reimplement pyvo, just extend it.

Unless it turns out that pyvo is not suitable at all, but I doubt it.

francoismg:
I have made the following changes and updates to the pyvo integration version :

- Added the possibility to choose a specific archive to query. The list of archives is loaded from the file in tool-data that is automatically generated through pyvo. The user can now choose to directly query the ivoa registry or choose a specific archive from the list
- The querying part has been reworked and the user can now choose between using no specific adql query (will fallback to standard ivoa.obscore query), create an ivoa.obscore table query with an interface containing the table field or create a custom query by choosing a specific table, where clause and the field that contains the url to use to download the file
- The tool now fails somewhat gracefully on most of the problems that can arise and generate a simple output stating that no files were found for these parameters

volodymyrss:
Thanks, it sounds interesting, could you please push it somewhere? Or did I miss it, I do not see the commits.

francoismg:
> Thanks, it sounds interesting, could you please push it somewhere? Or did I miss it, I do not see the commits.

yea it's pushed, you don't see it in the commits section?

francoismg:
> > > Thanks, it sounds interesting, could you please push it somewhere? Or did I miss it, I do not see the commits.
> > 
> > 
> > yea it's pushed, you don't see it in the commits section?
> 
> Ah, it did not show for me, you switched at some point from `tools-archives` to `main`. Could you please next time make a PR from a branch? It makes for a better git flow.
> 
> I will look in as is now.

oh ok sorry that's weird don't know how it switched to main, I made the initial pr from a new branch so it's weird, will look into it

PR #25, opened 2023-08-02
Archives tools prototypes

francoismg:
Tools prototypes for archives browsing


tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,164 @@
...
+    <command>
+      <![CDATA[
+
+        mkdir fits;
+
+        #if $query_section.query_selection.query_type == 'none':
+
+          #if $archiveselection.archive_type == 'registry' :
+
+            python '$__tool_directory__/archivespyvo.py' '$output' '$file_type' $number_of_files '$archiveselection.archive_type' '$archiveselection.keyword' '$archiveselection.wavebands' '$archiveselection.service_type' '$query_section.query_selection.query_type'
```
volodymyrss, line 19: where does `archivespyvo.py` come from? It's not in the PR. Maybe it's an old name of the script?

francoismg: oh yea sorry forgot to change it, it's the name I used in my install to not conflict with the other ones I made.
I will push the change with the correct name and you can just rename it archive.py in the meantime if you want to try it

volodymyrss: I thought so, but it had to be fixed anyway, thanks.

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,164 @@
...
+    </inputs>
+    <outputs>
+
+
+        <data name="output" format="txt" label="${tool.name} -> ${archiveselection.archive_type}:" >
+          <discover_datasets pattern="__designation_and_ext__" directory="fits" visible="true" format="fits" />
+        </data>
+
+    </outputs>
+    <help><![CDATA[
```
volodymyrss, line 148: seems like you still did you add tests. Could you please?
But anyway let's discuss tomorrow.

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,164 @@
...
+Archives description :
```
volodymyrss, line 149: just for the record, the help does not look right in galaxy. It is however a  priority at a later time.

image

volodymyrss:
- [ ] need "preview urls" option for output, using HTML with data previews etc
- [ ] need tests of 3 kinds:
   - [ ] non-strict checking of file properties (fits format, non-empty)
   - [ ] selecting nearly deterministic requests (e.g. run id, time range, see parameters)
   - [ ] use sitecustomize or tool test mode to return offline pre-stored remote responses
- [ ] need tool help with example
- [ ] need to complete all the labels and help for parameters 
- [ ] add citations

volodymyrss:
- [ ] need to have list of product types from ontology

tools/archives/esap_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,95 @@
+<tool id="astronomical_archives" name="Astronomical archives" version="1.0.0">
```
volodymyrss, line 1: I think we should remove this tool from this PR and keep it in a separate future PR.

volodymyrss:
just to the record, we might end up using an interactive tool in the future, e.g. like #26

francoismg:
> @francoismg I put some checkboxes. Could you see that it makes sense? Please ask questions if unclear or there are any roadblocks!

Yes everything seem clear right now, I'm just not sure about what you want to add in the citations section. 

Also the for the tap explorer I'm not sure that we can display something like this directly from a static file but I will try it and let you know.

francoismg:
> just to the record, we might end up using an interactive tool in the future, e.g. like #26

yea sure no problem, I thought it was already decided that it would probably be an interactive tool

volodymyrss:
> > just to the record, we might end up using an interactive tool in the future, e.g. like #26
> 
> yea sure no problem, I thought it was already decided that it would probably be an interactive tool

We might be talking about different things. Let's discuss later. But for now, let's just finish this first basic but already useful tool and find out with upstream developers how to best fill the gaps.

volodymyrss:
> > @francoismg I put some checkboxes. Could you see that it makes sense? Please ask questions if unclear or there are any roadblocks!
> 
> Yes everything seem clear right now, I'm just not sure about what you want to add in the citations section.

we want several things here, specifically:

10.3847/1538-3881/aabc4f

etc from

https://github.com/astropy/astropy/blob/main/astropy/CITATION

for  https://github.com/astropy/pyvo I do not see a DOI directly, just put the URL then

for https://www.ivoa.net/documents/latest/ADQL.html 

10.5479/ADS/bib/2008ivoa.spec.1030O

please put these and also a TODO to let me think more what to add

> Also the for the tap explorer I'm not sure that we can display something like this directly from a static file but I will try it and let you know.

Actually as discussed let's postpone this tap explored until after delivery of the first tool version. I will indicate this in the list.

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,954 @@
...
+-----
+
+The Planetary web services @ FUB's TAP end point. The Table Access
+Protocol (TAP) lets you execute queries against our database tables,
+inspect various metadata, and upload your own data. It is thus the
+VO's premier way to access public data holdings.
+
+Tables exposed through this endpoint include: columns, groups, key_columns, keys, schemas, tables from the tap_schema schema, epn_core from the hrsc3nd schema, emptyobscore, obscore from the ivoa schema.
+
+    </help>
```
volodymyrss, line 953:
```suggestion
    </help>
    <citations>
        <citation type="doi">
            10.3847/1538-4357/ac7c74
        </citation>
        <citation type="doi">
            10.3847/1538-3881/aabc4f
        </citation>
        <citation type="doi">
            10.5479/ADS/bib/2008ivoa.spec.1030O
        </citation>
        <citation type="doi">
            10.1051/0004-6361/201322068
        </citation>
        <citation type="bibtex">
        @ARTICLE{ivoa_arch,
           author = {Patrick Dowler and Janet Evans and Christophe Arviset and Severin Gaudet and Technical Coordination Group},
           title = {IVOA Architecture},
           url = {https://www.ivoa.net/documents/IVOAArchitecture/20211101/index.html},
           year = {2021},
           }
        </citation>
    </citations>
```

volodymyrss: I hope this represents citations well. We might need to adapt if some info is missing according to galaxy and doi resolver.

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+              <option value="none">No specific query (first n files from archive)</option>
+              <option value="obscore_query">IVOA obscore table query builder</option>
+              <option value="raw_query">Raw ADQL query builder</option>
+            </param>
+            <when value="obscore_query">
+              <param name="target_name" type="text" label="Target name" help="Name of the astronomical object observed" />
+              <param name="t_min" type="text" label="Start time in MJD" />
+              <param name="t_max" type="text" label="Stop time in MJD" />
+              <param name="dataproduct_type" type="text" label="Data product type" help="Logical data product type (image etc.)" />
+              <param name="dataproduct_type_select" type="select" label="Data product type" help="Logical data product type">
```
volodymyrss, line 103: please add reference where this comes from

tools/archives/pyvo_integration/astronomical_archives.py
```diff
@@ -0,0 +1,922 @@
...
+            access_url = 'access_url'
+
+        fits_file = FileHandler.download_file_from_url(file_url[0][access_url])
+        FileHandler.write_file_to_output(fits_file, output, "wb")
+
+        for i, url in enumerate(file_url[1:], start=1):
+            try:
+                fits_file = FileHandler.download_file_from_url(url[access_url])
+                FileHandler.write_file_to_subdir(fits_file, FileHandler.get_file_name_from_url(url[access_url]))
+            except Exception as e:
```
volodymyrss, line 888: handle this exception with some extra output to the user

tools/archives/pyvo_integration/astronomical_archives.py
```diff
@@ -0,0 +1,922 @@
...
+
+    if file_url and output != 'XXXX':
+
+        if query_type == 'raw_query':
+            access_url = url_field
+        else:
+            access_url = 'access_url'
+
+        fits_file = FileHandler.download_file_from_url(file_url[0][access_url])
+        FileHandler.write_file_to_output(fits_file, output, "wb")
```
volodymyrss, line 882: portentially this fails?

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+        </data>
+        <data name="output_html" format="html" label="${tool.name} -> HTML Preview:">
+          <filter>html_file == True</filter>
+        </data>
+        <data name="output_basic_html" format="html" label="${tool.name} -> Basic HTML Preview:">
+          <filter>basic_html_file == True</filter>
+        </data>
+        <data name="output_error" format="txt" label="${tool.name} -> Error Log:"></data>
+    </outputs>
+    <tests>
```
volodymyrss, line 183: add tests with failure, empty archive, no obscore, etc

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+              <param name="table" type="text" label="Table name" help="Name of the table you want to query (FROM clause)" required="True" />
+              <section name="where_clause" title="WHERE clause (field = value)" expanded="false">
+                <param name="where_field" type="text" label="WHERE field" help="WHERE **field** = value" />
+                <param name="where_condition" type="text" label="WHERE value" help="WHERE field = **value**" />
+              </section>
+              <param name="url_field" type="text" label="Url field" help="Table field containing the url of the file to download" required="True" />
+            </when>
+          </conditional>
+        </section>
+        <param name="number_of_files" type="integer" value="1" label="Number of files or urls to download" help="Min = 1 Max = 10" />
```
volodymyrss, line 161:
```suggestion
        <param name="number_of_files" type="integer" value="1" label="Number of files or urls to download" help="Min = 1 Max = 100; Beware of disk space usage when downloading large number of files!" />
```

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
+<tool id="astronomical_archives" name="Astronomical archives" version="1.0.0">
```
volodymyrss, line 1:
```suggestion
<tool id="astronomical_archives" name="Astronomical Archives (IVOA)" version="1.0.0">
```

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+    <description>queries astronomical archives</description>
```
volodymyrss, line 2:
```suggestion
    <description>queries astronomical archives through Virtual Observatory protocols</description>
```

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+    </command>
+    <inputs>
+        <conditional name="archiveselection">
+            <param name="archive_type" type="select" label="Archive Selection">
+              <option value="registry">Query IVOA registry</option>
+              <option value="archive">Query specific IVOA archive</option>
+            </param>
+            <when value="registry">
+              <param name="keyword" type="text" label="Keyword" />
+              <param name="service_type" type="select" label="Service type">
```
volodymyrss, line 64: please put a reference for versionned service type vocabulary. also for all other select options which rely on a particular version of some vocabularies

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+        </conditional>
+        <section name="query_section" title="Query selection" expanded="false">
+          <conditional name="query_selection">
+            <param name="query_type" type="select" label="ADQL Query Selection">
+              <option value="none">No specific query (first n files from archive)</option>
+              <option value="obscore_query">IVOA obscore table query builder</option>
+              <option value="raw_query">Raw ADQL query builder</option>
+            </param>
+            <when value="obscore_query">
+              <param name="target_name" type="text" label="Target name" help="Name of the astronomical object observed" />
```
volodymyrss, line 99:
```suggestion
              <param name="target_name" type="text" label="Observation target name" help="Typically name of the astronomical object observed" />
```

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+              <param name="t_min" type="text" label="Start time in MJD" />
+              <param name="t_max" type="text" label="Stop time in MJD" />
+              <param name="dataproduct_type" type="text" label="Data product type" help="Logical data product type (image etc.)" />
```
volodymyrss, line 102:
```suggestion
              <param name="t_min" type="float" label="Start time in MJD" />
              <param name="t_max" type="float" label="Stop time in MJD" />
              <param name="dataproduct_type" type="text" label="Data product type" help="Logical data product type (image etc.)" />
```

tools/archives/pyvo_integration/astronomical_archives.xml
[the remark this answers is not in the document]
volodymyrss: I'd suggest maybe to put select first, and if select is set to "custom", then reveal custom input test field.
Unless there is a "select + custom" parameter of course.

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+              <param name="dataproduct_type_select" type="select" label="Data product type" help="Logical data product type">
+                <option value="none" selected="true">None</option>
+                <option value="image">Image</option>
+                <option value="cube">Cube</option>
+                <option value="dynamic-spectrum">Dynamic spectrum</option>
+                <option value="event-list">Event list</option>
+                <option value="light-curve">Light curve</option>
+                <option value="measurements">Measurements</option>
+                <option value="polarization-cube">Polarization cube</option>
```
volodymyrss, line 111: these two have the same value. In fact, it looks like only first one is shown in UI. when you put the reference for the list, we can check what to do best.

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+                <option value="spectral-cube">Spectral cube</option>
+                <option value="spectrally-resolved-dataset">Spectrally resolved dataset</option>
+                <option value="spectrum">Spectrum</option>
+                <option value="temporally-resolved-dataset">Temporally resolved dataset</option>
+                <option value="time-cube">Time cube</option>
+                <option value="timeseries">Time series</option>
+                <option value="velocity-curve">Velocity curve</option>
+                <option value="visibility">Visibility</option>
+              </param>
+              <param name="obs_collection" type="text" label="Collection" help="Name of the data collection" />
```
volodymyrss, line 126:
```suggestion
              <param name="obs_collection" type="text" label="Collection" help="Name of the observation collection" />
```

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+              <param name="obs_title" type="text" label="Title" help="Title" />
```
volodymyrss, line 127:
```suggestion
              <param name="obs_title" type="text" label="Title" help="Observation title (archive-specific)" />
```

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+              <param name="obs_id" type="text" label="Id" help="Id" />
```
volodymyrss, line 128:
```suggestion
              <param name="obs_id" type="text" label="Observation Id" />
```

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+              <param name="facility_name" type="text" label="Facility name" help="Name of the facility used for this observation" />
+              <param name="instrument_name" type="text" label="Instrument name" help="Name of the instrument used for this observation" />
+              <param name="em_min" type="text" label="Spectral coordinates (start)" />
+              <param name="em_max" type="text" label="Spectral coordinates (end)" />
+              <param name="obs_publisher_id" type="text" label="Publisher dataset ID" />
```
volodymyrss, line 133:
```suggestion
              <param name="em_min" type="float" label="Minimum Energy (band start in spectral coordinates), archive-specific unit" />
              <param name="em_min" type="float" label="Maximal Energy (band stop in spectral coordinates), archive-specific unit" />
              <param name="obs_publisher_id" type="text" label="Publisher dataset ID" />
```

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+                <option value="1">1</option>
+                <option value="2">2</option>
+                <option value="3">3</option>
+                <option value="4">4</option>
+                <option value="5">5</option>
+              </param>
+              <param name="order_by" type="select" label="Order by" help="Field to use for ordering the results">
+                <option value="none" selected="true">No specific order</option>
+                <option value="size" >Order by size (Ko)</option>
+                <option value="collection" >Order by collection</option>
```
volodymyrss, line 147: collection name, alphabetically?

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+                <option value="object" >Order by name</option>
```
volodymyrss, line 148: by object name, alphabetically?

francoismg: yes default is ascend alphabetically

tools/archives/pyvo_integration/astronomical_archives.xml
francoismg: I'm looking for the reference right now will put it as soon as I find it

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+            </when>
+        </conditional>
+        <section name="query_section" title="Query selection" expanded="false">
+          <conditional name="query_selection">
+            <param name="query_type" type="select" label="ADQL Query Selection">
+              <option value="none">No specific query (first n files from archive)</option>
+              <option value="obscore_query">IVOA obscore table query builder</option>
+              <option value="raw_query">Raw ADQL query builder</option>
+            </param>
+            <when value="obscore_query">
```
volodymyrss, line 98: Please check if this is what you used

```suggestion
            <when value="obscore_query">
            <!-- Fields defined https://www.ivoa.net/documents/ObsCore/20160224/WD-ObsCore-v1.1-20160224.pdf -- >
```

francoismg: I was kooking at it right now, it's not the same document but the tables content looks more or less the same

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+                <param name="where_condition" type="text" label="WHERE value" help="WHERE field = **value**" />
+              </section>
+              <param name="url_field" type="text" label="Url field" help="Table field containing the url of the file to download" required="True" />
+            </when>
+          </conditional>
+        </section>
+        <param name="number_of_files" type="integer" value="1" label="Number of files or urls to download" help="Min = 1 Max = 10" />
+        <param name="csv_file" type="boolean" truevalue="true" falsevalue="false" label="Download urls as csv"/>
+        <param name="image_file" type="boolean" truevalue="true" falsevalue="false" label="Download files"/>
+        <param name="html_file" type="boolean" truevalue="true" falsevalue="false" label="Preview urls in extended html"/>
```
volodymyrss, line 164:
```suggestion
        <param name="html_file" type="boolean" truevalue="true" falsevalue="false" label="Preview URLs in extended HTML"/>
```

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+        <param name="basic_html_file" type="boolean" truevalue="true" falsevalue="false" label="Preview urls in basic html"/>
```
volodymyrss, line 165:
```suggestion
        <param name="basic_html_file" type="boolean" truevalue="true" falsevalue="false" label="Preview URLs as HTML"/>
```

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1005 @@
...
+        <data name="output_csv" format="txt" label="${tool.name} -> CSV Manifest:">
+          <filter>csv_file == True</filter>
+        </data>
+        <data name="output_html" format="html" label="${tool.name} -> HTML Preview:">
+          <filter>html_file == True</filter>
+        </data>
+        <data name="output_basic_html" format="html" label="${tool.name} -> Basic HTML Preview:">
+          <filter>basic_html_file == True</filter>
+        </data>
+        <data name="output_error" format="txt" label="${tool.name} -> Error Log:"></data>
```
volodymyrss, line 181: It's not nice that it shows "Error Log" at start of every request.

It may be that it's best to just write to `stdout` and then see it in tool execution if needed.

Or/and, we could do like this:

```suggestion
        <data name="output_error" format="txt" label="${tool.name} -> Query summary:"></data>
```

francoismg: Goal was to provide user with more information about why the query might have failed but yes Query summary sounds better and I can also write the error to stdout in addition no problem

volodymyrss: If proposed change is ok for you, you can just `commit suggestion` without pushing anything else. I will just do it in this case, but see other suggestions.

tools/archives/pyvo_integration/astronomical_archives.xml
francoismg: > these two have the same value. In fact, it looks like only first one is shown in UI. when you put the reference for the list, we can check what to do best.

finally found the list I used https://www.ivoa.net/rdf/product-type/2023-06-26/product-type.html#image

the double polarization-cube was an error from my part should have been "polarization-resolved-dataset". I have changed it in the code already

volodymyrss: > > these two have the same value. In fact, it looks like only first one is shown in UI. when you put the reference for the list, we can check what to do best.
> 
> finally found the list I used https://www.ivoa.net/rdf/product-type/2023-06-26/product-type.html#image

That's a good one, definitely put a reference to it. Note that there is also machine-readable rdf there which in principle can be used to generate the list you have here.

volodymyrss: But also put a reference to the pdf.

volodymyrss:
You said there is a version which passes tests now? @francoismg ?

volodymyrss:
> > You said there is a version which passes tests now? @francoismg ?
> 
> yes all the functionnal tests pass now

But it's not pushed?

francoismg:
One linting error for xml tool (unreachable url for VizieR) in the help section -> will strip the url so it won't be tested as a url

Some linting errors in the python script -> mostly too long lines, import order to change and 2 unused variables

Tool test (with planemo test) seems successfull but apparently tests have failed when looking at the detail (they pass when I'm running planemo myself so between that and the fact that the global test is green maybe it's good)

"Combined chunked test results" seems to fail because of the results of the others tests maybe will pass after linting is fixed

"Check workflow success" seems to fail because of the linting fail so will probably pass after linting is fixed

volodymyrss:
> One linting error for xml tool (unreachable url for VizieR) in the help section -> will strip the url so it won't be tested as a url
> Some linting errors in the python script -> mostly too long lines, import order to change and 2 unused variables

Ok, sounds good

> Tool test (with planemo test) seems successfull but apparently tests have failed when looking at the detail (they pass when I'm running planemo myself so between that and the fact that the global test is green maybe it's good)
> 
> "Combined chunked test results" seems to fail because of the results of the others tests maybe will pass after linting is fixed
> 
> "Check workflow success" seems to fail because of the linting fail so will probably pass after linting is fixed

Ok, I see. Let's fix as much as possible and propose this soon, hopefully today.

francoismg:
I saw that the "fail_level" parameter was set to warn in their workflow, instead of error, so every warning must be fixed instead of error only. Right now I think everything has been fixed but I had to switch from the static file to a data table, which is working expect that it returns id value instead of the archive url so I'm still trying to figure that out.

- Every python linting errors have been fixed (probably more since I got more things when running "lint_shed" locally)
- Every tool linting errors and warnings have been fixed (except for the unexisting "Astronomy" category in tool shed and the data table thing I mentioned)

Think the data table problem should be good soon, should be able to push again before noon hopefully

tools/archives/pyvo_integration/tool-data/astronomical_archives_gen.loc
```diff
@@ -0,0 +1,139 @@
+# Table used for listing astronomical archives TAP service urls
+#<id>	<display_name>	<value>\n0	AIP DaCHS TAP service	http://gavo.aip.de/tap
```
volodymyrss, line 2: this `\n` before line 0 seems suspicious

francoismg: yea looks like a space has been skipped during generation of the file, it doesn't seem to cause bugs except for the fact that this archive one is not available on the list. Fix has been pushed.

But I don't think it's related to the failed tests, when I run the thing locally with planemo test or serve everything works fine. The data table is implemented a bit differently whether  the tool is directly inside galaxy or from the tool shed but both worked when I tested it and the tests failed the same way before any changes.

The problem seems to be docker related, exit code 125 so I don't know if it's a tool problem or a problem with the workflow. 

I tried to look at planemo source and search if I found other people with same problem but have found nothing so far.

I looked at the logs from your cta and there's no error message after the container is instantiated

volodymyrss: For the error message, it's a separate point indeed, let's ask developers after this is ready.

francoismg: That's the only problem remaining, everything else seems to pass now. Maybe the container failling is not a problem in itself but the fact that it writes to the error stream makes galaxy think tool run has failed, will try to change how the tool react to something written to error stream.

Did you update the workflow between now and your latest cta tool test?

volodymyrss:
Hi @bgruening , @francoismg made here in coordination with @andriineronov  and myself some tool to explore and fetch data from IVOA (Virtual Observatory) protocols, what we discussed on matrix in June. It also addresses question many astronomers have about Galaxy - how to get our data into it? (upload is not an option)

In principle, it's doing what we want and is actually useful, just one screenshot:

image

However we are facing one essential challenge:

* we'd like to make some elements of input and output dynamic, there are two related aspects to this
  * some input parameters should have dynamically generated options: e.g. archive list comes from IVOA archive registry
  * now, by default, tool output includes only list of data products. we look on them, decide better criteria, and then re-run the tool with "download" option. But we'd want that we have UI which allows user to directly select products from tool output and run new tool with them. 
 
What is the best way to do this?  I imagine we could split this process in two sequential jobs, but it becomes less interactive. Also I do not think we can pick files from nice HTML output with helpful metadata (see screenshot), and if it's a lot of files. 

smaller issues:

* tests fail for unknown to us reasons, possibly not due to our tool
* I find that passing parameters from galaxy runner to python script we have here is not totally ideal as it is prone to errors with all the positional parameters. Is it possible maybe to pass entire parameter set as json, or something?

francoismg:
@volodymyrss regarding the parameters part, goal was to use something like https://docs.python.org/3/library/argparse.html so we can have changing numbers of named parameters. That way we can have only one script call too since we could generate the call on the fly.

Tried it a bit at first but had some parsing error so I used the standard way in the meantime but will look into it tomorrow

bgruening:
> * I find that passing parameters from galaxy runner to python script we have here is not totally ideal as it is prone to errors with all the positional parameters. Is it possible maybe to pass entire parameter set as json, or something?

Please have a look at https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-configfiles-configfile , the entire section please.
You can have something like:

```xml
<configfiles>
    <inputs name="inputs" filename="inputs.json" />
</configfiles>
```

I prefer this more explicit here: https://github.com/galaxyproject/galaxy/blob/02fb8c2b1d314310fc3be7ec048aff399321e9a5/test/functional/tools/tool_provided_metadata_2.xml#L8

bgruening:
> tests fail for unknown to us reasons, possibly not due to our tool

No, we see that since yesterday in many repo. Something changed, we are tracking it down. Sorry.

bgruening:
> However we are facing one essential challenge:
> 
>     * we'd like to make some elements of input and output dynamic, there are two related aspects to this
>       
>       * some input parameters should have dynamically generated options: e.g. archive list comes from IVOA archive registry
>       * now, by default, tool output includes only list of data products. we look on them, decide better criteria, and then re-run the tool with "download" option. But we'd want that we have UI which allows user to directly select products from tool output and run new tool with them.
> 
> 
> What is the best way to do this? I imagine we could split this process in two sequential jobs, but it becomes less interactive. Also, I do not think we can pick files from nice HTML output with helpful metadata (see screenshot), and if it's a lot of files.

Can you write a https://docs.pyfilesystem.org/en/latest/ plugin? Would it be possible, just architecture-wise, to implement the selection process in something like a fuse mount? Where your users iterate through virtual folders, select virtual files and then hit GO. If this is conceptually possible I would propose that you implement a pyfilesystem plugin for your "archives". This is independent and might be useful beyond Galaxy. Galaxy can then take this pyfilesystem plugin and give you a "file-browser" where people can select data.

Would that work?

If not, it would help to see a screenshot of the parameters. But I think that everything that is highly interactive should not be modeled with the tool framework. A [data-source](https://pubmed.ncbi.nlm.nih.gov/21531983/), maybe an [Interactive Tool](https://live.usegalaxy.eu/) are maybe more appropriate. 

Is there already a UI to filter, and select with the Archive? Somewhere?

volodymyrss:
> Please have a look at https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-configfiles-configfile , the entire section please. You can have something like:
> ```
> <configfiles>
>     <inputs name="inputs" filename="inputs.json" />
> </configfiles>
> ```

This seems to be exactly what I meant, thanks.
We have up to 24 input parameters, and they are substituted first into tool command, and then interpreted by python.
Instead, we can have galaxy input parameters directly inside python in json. You see, @francoismg ? Looks like we do not need any argument parsing (even with argparse). Just parse json inside python tool and do all the logic there. There is an [example tool](https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/inputs_as_json.xml) which does this in the doc.
 
> I prefer this more explicit here: https://github.com/galaxyproject/galaxy/blob/02fb8c2b1d314310fc3be7ec048aff399321e9a5/test/functional/tools/tool_provided_metadata_2.xml#L8

This I did not understand, sorry :( This example does not even have inputs?..

bgruening:
Is that one better: https://github.com/galaxyproject/galaxy/blob/02fb8c2b1d314310fc3be7ec048aff399321e9a5/test/functional/tools/column_param_configfile.xml#L7

Essentially you can write into a `configfile` whatever you like and have direct access to the input/output variables.
This means you can construct a json, or an entire python script on the fly. No argument parsing is needed.

volodymyrss:
> Is that one better: https://github.com/galaxyproject/galaxy/blob/02fb8c2b1d314310fc3be7ec048aff399321e9a5/test/functional/tools/column_param_configfile.xml#L7
> 
> Essentially you can write into a `configfile` whatever you like and have direct access to the input/output variables. This means you can construct a json, or an entire python script on the fly. No argument parsing is needed.

I dealt with this in [here](https://github.com/esg-epfl-apc/tools-astro/pull/22) too. Writing python file in xml directly prevents linting (or is there an adapted linted for python script within galaxy tool xml?), makes testing slower (by at least some 10 seconds to do planemo test). 

But I think it is understood that there is a reasonable limit on the size and complexity of the script to put into tool xml, no?  Also script might be from an external package, so arguments need to be passed to it, since the script can not be templated.

So, if we can not always template the whole script, we need to pass arguments to it.
If we have 24 arguments, how do we pass them? Positional is difficult, especially if meaning of position depends on previous positions.
We could add 24 `--argument-X=$argument_X` strings. And every time we change/add a tool argument, we'd need to add/change this passing. Also these strings seem obvious to construct automatically.
Putting all existing inputs, without listing them again, into `inputs.json` would seem to address this issue. Inside python tool we can even parse this json directly into argparse's structure.

Or am I missing some easier way you suggest?
It seems to me that keeping explicit config with all 24 values will mean adding extra effort to make sure this explicit config is compatible with both script and galaxy tool inputs, no?

bgruening:
I'm not at all arguing for argparse, use the configfile to produce a Json and pass that to your pthon code. I just lost options :-) you choose what you like best. :-)

volodymyrss:
> Can you write a https://docs.pyfilesystem.org/en/latest/ plugin? Would it be possible, just architecture-wise, to implement the selection process in something like a fuse mount? Where your users iterate through virtual folders, select virtual files and then hit GO. If this is conceptually possible I would propose that you implement a pyfilesystem plugin for your "archives". This is independent and might be useful beyond Galaxy. Galaxy can then take this pyfilesystem plugin and give you a "file-browser" where people can select data.
> Would that work?

The structure we have is more like relational database, with several tables and one principle table which returns "files" (data products) with a lot of columns. We need to see this columns to choose files. See screenshot above.

I do not currently see a way in pyfilesystem to show those. Maybe I missed it?

In fact, I wonder if it there is a generic way to access, visualize, select from SQL-like storage with galaxy. If there is, we could write an extension to that.

Folder structure is not really the way we (users) see these archives. Or rather, sometimes we do, but it is not really useful and not sufficient, since the folder structure typically made out of incomprehensible (for anyone but the very experts) observation IDs. In principle, we could add this just for the few data center support people like ourselves, but we do not really benefit from the galaxy as much as our users who do need it :)

Also, related, we need to search over this file collection, table, using the column values (like in SQL). We can see these values as file metadata.
We can not really search with such amount of detail in pyfilesystem I think, right? Since it has no concept of extended file metadata, as far as I can see, for viewing or searching.



> 
> If not, it would help to see a screenshot of the parameters. 

Here are some parameters:

image

here I made a video https://fair.tube/w/bAYTSeD7gt2XiDTWzTo3rz

> But I think that everything that is highly interactive should not be modeled with the tool framework. 

I am not sure yet it is unfeasible to make our case without interactive tool. It seems preferable to use tool framework, no?

But there are still couple of issues not exactly because of interactivity. Apart from lack (of our awareness) of SQL-like storage interface from galaxy, there is this possible difficulty:

The tool produces some values (e.g. first request to the tool found some telescopes which the user apparently wants data from). 
We need to use these values as inputs to the next execution of the tool. Currently we copy-paste values written in the output file.
Is there way to click somewhere and move the output of one tool to another?
I guess since output is file/collection, if we use file input for some selection this would work. I am not sure if there is an easier way. There is no way to output a float or a string? 

We show this HTML output file,  is it possible to click somewhere in it to start tool with pre-filled values?

> A [data-source](https://pubmed.ncbi.nlm.nih.gov/21531983/), maybe an [Interactive Tool](https://live.usegalaxy.eu/) are maybe more appropriate.

Thanks a lot, I add a comment to [our issue ](https://github.com/esg-epfl-apc/tools-astro/issues/26)
 
> Is there already a UI to filter, and select with the Archive? Somewhere?

So that's funny, there are numerous UI's (e.g. here is a [small one](http://voparis-srv.obspm.fr/portal/) by our colleagues, I could give many more examples) but there is not a common single one. These protocols are used extensively in various applications, UI every time adapted to particular case, particular platform.

The need to have a more universal sustainable and reusable UI is more and more recognized and is now part of architecture of currently built projects (it reminds me how Galaxy spread as a common data analysis platform UI).
As one of the effects of this, here is a prototype UI (called ESAP) developed by ESCAPE (recently completed EU project), this is our test deployment:

https://esap-gui.test-cta-cscs.odahub.io/
https://fair.tube/w/8ZztBmpX53Jzp7v9fCZuep

As you can see it is a bit rough, but it shows all the right pieces in the right order, how astronomers expect them.
One of the problems, as always, is sustainability, and this is being figured out at many levels.

That's why I think that if EuroScienceGateway deploys an astronomer-friendly UI for IVOA archives in usegalaxy.eu this 
could be very interesting since it will live there for a while, and astronomers will go there as a single  understandable place to fetch data from IVOA archives.

We think even what we have here is not too bad already, and we can start showing it. Do you think it makes sens to converge on the tool as it is soon (I will make a talk on ESG on Swiss SKA days in two weeks, they are very interested in this in particular).
And then identify the way to address these issues we have?

volodymyrss:
> I'm not at all arguing for argparse, use the configfile to produce a Json and pass that to your pthon code. I just lost options :-) you choose what you like best. :-)

Just wanted to make sure we get all the information we need to make a choice, maybe I missed something! :) This will work then.

bgruening:
The tests should work now again.

volodymyrss:
> The tests should work now again.

Something still fails for us it seems.

bgruening:
Can you give me access to this repo? I will change some CI stuff in this PR here.

bgruening:
Back to the actual use-case problem. Long story short, I don't think we do have the perfect solution at hand at the moment.

## Tool-framework
Tools are on purpose not interactive, tools should look and behave similar to users, so they can easily switch between all 3000 tools. They are submitting jobs to a cluster, which has a certain overhead and I guess we would need to have 2-3 tools. It will work but will feel clunky.

## Interactive tools framework 
Super flexible. You could use one of the existing GUIs and sent the final URLs to the upload tool in Galaxy from the IT.
But this needs to spawn a Docker container before you see a GUI. So not instantaneous.

## Visualisation framework

Super flexible, instantaneous, client-side only. Does not need a container, probably best UX.
But its a "visualization" and will be hard to discover for user as those Vis are not part of the tool panel. We would need to come up with an idea how to get this "Visualisation" into the Tool-Panel.
This will work today, because the JS Visualistion can access Galaxy API and sent URL to the upload tools, it has probably the best UX or your use-case - downside is the find-ability of this Visualisation/Tool hybrid.

## New tool parameter

We could think about adding a new tool-parameter. One that executes SQL against an endpoint. I don't know if we can design one that is generic enough to be honest. But @annefou has a similar use-case and maybe we can come up with a proposal? It would take longer but is maybe the better long-term plan?


Does that help to choose a way forward?

bgruening:
If you have chosen your preferred way, please let me know and I can give you maybe pointers for similar stuff hat already exists.

bgruening:
> We think even what we have here is not too bad already, and we can start showing it. Do you think it makes sens to converge on the tool as it is soon (I will make a talk on ESG on Swiss SKA days in two weeks, they are very interested in this in particular).

Yes, I do think what you have is already nice. And I'm happy to deploy it and seek for more feedback, we can always improve UX as we go. I'm pragmatic and whatever helps our users, even if its not perfect, is a win for users.

bgruening:
> https://esap-gui.test-cta-cscs.odahub.io/
https://fair.tube/w/8ZztBmpX53Jzp7v9fCZuep

If you can make changes to it and deploy it in a Docker container, we could start it via ITs (similar to Juypter Notebooks) and add a function to the GUI (of this webapp) to directly upload via the user-API key data to the history. But its not instantanous. See my post above.

Thanks for sharing all that, this was very helpful for me to understand this usecase.

volodymyrss:
> > We think even what we have here is not too bad already, and we can start showing it. Do you think it makes sens to converge on the tool as it is soon (I will make a talk on ESG on Swiss SKA days in two weeks, they are very interested in this in particular).
> 
> Yes, I do think what you have is already nice. And I'm happy to deploy it and seek for more feedback, we can always improve UX as we go. I'm pragmatic and whatever helps our users, even if its not perfect, is a win for users.

I will show it on [SKA days](https://indico.skatelescope.org/event/1031/contributions/9848/)  in two weeks

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1023 @@
...
+        <edam_operation>operation_0224</edam_operation>
+    </edam_operations>
+    <requirements>
+        <requirement type="package" version="5.2.2">astropy</requirement>
+        <requirement type="package" version="1.4.1">pyvo</requirement>
+    </requirements>
+    <command>
+      <![CDATA[
+
+        mkdir fits;
```
bgruening, line 13: its better to use && instead of ; ... && fails if one command fails

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1023 @@
...
+            <param name="query_type" type="select" label="ADQL Query Selection">
+              <option value="none">No specific query (first n files from archive)</option>
+              <option value="obscore_query">IVOA obscore table query builder</option>
+              <option value="raw_query">Raw ADQL query builder</option>
+            </param>
+            <when value="none"></when>
+            <when value="obscore_query">
+              <!-- Fields defined https://www.ivoa.net/documents/ObsCore/20160224/WD-ObsCore-v1.1-20160224.pdf -->
+              <param name="target_name" type="text" label="Observation target name" help="Typically name of the astronomical object observed" />
+              <param name="t_min" type="float" optional="true" label="Start time in MJD" />
```
bgruening, line 105: can we add min/max?

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1023 @@
...
+          <conditional name="query_selection">
+            <param name="query_type" type="select" label="ADQL Query Selection">
+              <option value="none">No specific query (first n files from archive)</option>
+              <option value="obscore_query">IVOA obscore table query builder</option>
+              <option value="raw_query">Raw ADQL query builder</option>
+            </param>
+            <when value="none"></when>
+            <when value="obscore_query">
+              <!-- Fields defined https://www.ivoa.net/documents/ObsCore/20160224/WD-ObsCore-v1.1-20160224.pdf -->
+              <param name="target_name" type="text" label="Observation target name" help="Typically name of the astronomical object observed" />
```
bgruening, line 104: can we help here with a select list?

bgruening: we could add a select list with the most common objects and then offer a free text option?

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1023 @@
...
+              <param name="t_max" type="float" optional="true" label="Stop time in MJD" />
+              <!-- Dataproduct_type field values definition https://www.ivoa.net/rdf/product-type/2023-06-26/product-type.html  -->
+              <conditional name="dataproduct_type_selection">
+                <param name="dataproduct_input_type" type="select" label="Data product type input choice">
+                  <option value="list" selected="true">From list</option>
+                  <option value="custom">Custom (Input data product type yourself)</option>
+                </param>
+                <when value="list">
+                  <param name="dataproduct_type_select" type="select" label="Data product type" help="Logical data product type (link to reference in Help section)">
+                    <option value="none" selected="true">None</option>
```
bgruening, line 115: This None things we can omit by simply add `optional="true"` to the select.

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1023 @@
...
+                    <option value="timeseries">Time series</option>
+                    <option value="velocity-curve">Velocity curve</option>
+                    <option value="visibility">Visibility</option>
+                  </param>
+                </when>
+                <when value="custom">
+                  <param name="dataproduct_type" type="text" label="Data product type" help="Logical data product type" />
+                </when>
+              </conditional>
+              <param name="obs_collection" type="text" label="Collection" help="Name of the observation collection" />
```
bgruening, line 143: I guess the description could be the label and you can drop the help

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1023 @@
...
+              <param name="obs_title" type="text" label="Observation title (archive specific)" />
+              <param name="obs_id" type="text" label="Observation Id" />
+              <param name="facility_name" type="text" label="Facility name" help="Name of the facility used for this observation" />
+              <param name="instrument_name" type="text" label="Instrument name" help="Name of the instrument used for this observation" />
+              <param name="em_min" type="text" label="Start of the energy range, vacuum wavelength in meters" />
```
bgruening, line 148: in meters, indicates and int/float?

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1023 @@
...
+              <param name="em_max" type="text" label="Stop of the energy range, vacuum wavelength in meters" />
+              <param name="obs_publisher_id" type="text" label="Publisher dataset ID" />
+              <param name="s_fov" type="text" label="Diameter (bounds) of the covered region (deg)" />
+              <param name="calibration_level" type="select" label="Calibration level (0, 1, 2, 3, 4, 5)" >
```
bgruening, line 152: why not an integer with optional="true" and min=0 and max=5? It would give you a nice slider.

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1023 @@
...
+                <option value="none">None</option>
+                <option value="0">0</option>
+                <option value="1">1</option>
+                <option value="2">2</option>
+                <option value="3">3</option>
+                <option value="4">4</option>
+                <option value="5">5</option>
+              </param>
+              <param name="order_by" type="select" label="Order by" help="Field to use for ordering the results">
```
bgruening, line 161: optional="true"

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1023 @@
...
+                <option value="object" >Order by Target Name, alphabetically</option>
+              </param>
+            </when>
+            <when value="raw_query">
+              <param name="table" type="text" label="Table name" help="Name of the table you want to query (FROM clause)" />
+              <section name="where_clause" title="WHERE clause (field = value)" expanded="false">
+                <param name="where_field" type="text" label="WHERE field" help="WHERE **field** = value" />
+                <param name="where_condition" type="text" label="WHERE value" help="WHERE field = **value**" />
+              </section>
+              <param name="url_field" type="text" label="Url field" help="Table field containing the url of the file to download" />
```
bgruening, line 174: does this need https:// or not?

volodymyrss:
> > https://esap-gui.test-cta-cscs.odahub.io/
> > https://fair.tube/w/8ZztBmpX53Jzp7v9fCZuep
> 
> If you can make changes to it and deploy it in a Docker container, we could start it via ITs (similar to Juypter Notebooks) and add a function to the GUI (of this webapp) to directly upload via the user-API key data to the history. But its not instantanous. See my post above.

It's two containers actually (frontend and backend), IT supports this?

This is an interesting option since it uses synergy with past project, but it also depends on sustainability of ESAP, unless we want to fully support our fork.

But, if we go for IT option, we could certainly reuse part of ESAP (it has a lot of what we do not need for this feature here).

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1023 @@
...
+            </when>
+          </conditional>
+        </section>
+        <param name="number_of_files" type="integer" value="1" label="Number of files or urls to download" help="Min = 1 Max = 100; Beware of disk space usage when downloading large number of files!" />
```
bgruening, line 178:
```suggestion
        <param name="number_of_files" type="integer" value="1" min="1" max="100" label="Number of files or urls to download" help="Beware of disk space usage when downloading large number of files!" />
```

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1023 @@
...
+        <param name="csv_file" type="boolean" truevalue="true" falsevalue="false" label="Download URLs as CSV"/>
```
bgruening, line 179: this is an additional download, additional to FITS, correct? maybe mention that?

volodymyrss:
> Back to the actual use-case problem. Long story short, I don't think we do have the perfect solution at hand at the moment.
> ## Tool-framework
> 
> Tools are on purpose not interactive, tools should look and behave similar to users, so they can easily switch between all 3000 tools. They are submitting jobs to a cluster, which has a certain overhead and I guess we would need to have 2-3 tools. It will work but will feel clunky.

It seems that we can make several tools as part of other solutions too, right? So we continue doing this anyway.
It's good that this allows to re-run, export workflow, publish, right? That's important.

> ## Interactive tools framework
> 
> Super flexible. You could use one of the existing GUIs and sent the final URLs to the upload tool in Galaxy from the IT. But this needs to spawn a Docker container before you see a GUI. So not instantaneous.

Can the IT container be pre-spawned? Kept spawned for a while?
What happens to the history in this case, I guess it's not possible to automatically re-run IT tool getting the same outcome, right?
If so, the only advantage comparing to Visualization is that it's easier to find in UI?

> ## Visualisation framework
> 
> Super flexible, instantaneous, client-side only. Does not need a container, probably best UX. But its a "visualization" and will be hard to discover for user as those Vis are not part of the tool panel. We would need to come up with an idea how to get this "Visualisation" into the Tool-Panel. This will work today, because the JS Visualistion can access Galaxy API and sent URL to the upload tools, it has probably the best UX or your use-case - downside is the find-ability of this Visualisation/Tool hybrid.

Indeed this seems best. Can we not put a link in some obvious place, like in top menu? Just in some instances/forks?

> ## New tool parameter
> 
> We could think about adding a new tool-parameter. One that executes SQL against an endpoint. I don't know if we can design one that is generic enough to be honest. But @annefou has a similar use-case and maybe we can come up with a proposal? It would take longer but is maybe the better long-term plan?

Ok let's discuss this as a long-term plan, do we make an issue somewhere for planning and discussing?

bgruening:
There are now just two failing tests:

grafik

tools/archives/pyvo_integration/astronomical_archives.xml
volodymyrss, line 13: would this work too?  else we need to change every command followed by something.
Actually as @francoismg will change this treatment to json, it will be easy to do &&

```suggestion
        mkdir fits || exit 1
```

volodymyrss: would this work too?  else we need to change every command followed by something.
Actually as @francoismg will change this treatment to json, it will be easy to do &&

```suggestion
        mkdir fits || exit 1
```

tools/archives/pyvo_integration/astronomical_archives.xml
volodymyrss: Can we do both in one field, select+freetext? Our too superficial search did not show such an input.

tools/archives/pyvo_integration/astronomical_archives.xml
volodymyrss: Min is a bit arbitrary (0 is fine in most cases, it's 1858), and max should in principle depend on current moment, but I set now+10 years. We can not make it dynamic I guess?

```suggestion
              <param name="t_min" type="float" min=0 max=63832 optional="true" label="Start time in MJD" />
```

bgruening: dynamic? you mean dependent on another param? No that is not possible yet. If no min/max possible, then leave it like it is. min/max can just improve UX

tools/archives/pyvo_integration/astronomical_archives.xml
bgruening: Not on one, you would need a conditional like you used below.

tools/archives/pyvo_integration/astronomical_archives.xml
bgruening: Yes you can do that, but isn't that a bit too complicated :) && is kindof your recommendation.

bgruening:
> It's two containers actually (frontend and backend), IT supports this?

No, we would need to have one single container atm.

> It seems that we can make several tools as part of other solutions too, right? So we continue doing this anyway.
It's good that this allows to re-run, export workflow, publish, right? That's important.

Yes, its actually a good point. You can not use the Visualisation inside a workflow. But our recommendation is to keep data download seperate from workflows. Aka upload, then start your workflows.

> Indeed this seems best. Can we not put a link in some obvious place, like in top menu? Just in some instances/forks?

Yes, we need to come up with something. Its just not there yet. 

> Ok let's discuss this as a long-term plan, do we make an issue somewhere for planning and discussing?

https://github.com/galaxyproject/galaxy/issues/6806 but maybe a new issue is better suited and link to that issue.

tools/archives/pyvo_integration/astronomical_archives.xml
volodymyrss: After there are many options too add and two more commands so it needs to be added everywhere. Ok we'll see.

bgruening: every command should be separated by && to the next comment. I'm not sure I understand the problem. Maybe I'm missing something. Its really just replacing `;` with `&&`

tools/archives/pyvo_integration/astronomical_archives.xml
volodymyrss: It's current moment, so depending on current time.

tools/archives/pyvo_integration/astronomical_archives.xml
volodymyrss: What value will it have when it is optional?

tools/archives/pyvo_integration/astronomical_archives.xml
volodymyrss:
```suggestion
              <param name="obs_collection" type="text" label="Observation collection name"/>
```

tools/archives/pyvo_integration/astronomical_archives.xml
volodymyrss: There is no problem, it's just along command. It will become easier and then we replace, thanks!

tools/archives/pyvo_integration/astronomical_archives.xml
volodymyrss: In principle, would it be useful to have select-or-text input?
In select, one can search with text. Then, if not found, user has to copy what is inputed, click "input manually" and then input in another field?

tools/archives/pyvo_integration/astronomical_archives.xml
volodymyrss:
```suggestion
              <param name="em_min" type="float" label="Start of the energy range, vacuum wavelength in meters" />
```

tools/archives/pyvo_integration/astronomical_archives.xml
bgruening: Yes, I think its a useful idea. We should do it at some point.

tools/archives/pyvo_integration/astronomical_archives.xml
volodymyrss: It's name of the table column which contains URLs. Name of the column itself is a string with no http://

tools/archives/pyvo_integration/astronomical_archives.xml
volodymyrss: It's one of the options, just like downloading FITS.
```suggestion
        <param name="csv_file" type="boolean" truevalue="true" falsevalue="false" label="Return URL list as CSV"/>
```

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1023 @@
...
+                <param name="where_condition" type="text" label="WHERE value" help="WHERE field = **value**" />
+              </section>
+              <param name="url_field" type="text" label="Url field" help="Table field containing the url of the file to download" />
+            </when>
+          </conditional>
+        </section>
+        <param name="number_of_files" type="integer" value="1" min="1" max="100" label="Number of files or urls to download" help="Beware of disk space usage when downloading large number of files!" />
+        <param name="csv_file" type="boolean" truevalue="true" falsevalue="false" label="Download URLs as CSV"/>
+        <param name="image_file" type="boolean" truevalue="true" falsevalue="false" label="Download files"/>
+        <param name="html_file" type="boolean" truevalue="true" falsevalue="false" label="Preview URLs in extended HTML (requires HTML rendering permission, see help)"/>
```
volodymyrss, line 181:
```suggestion
        <param name="html_file" type="boolean" truevalue="true" falsevalue="false" label="Return URL list in extended HTML (requires HTML rendering permission, see help)"/>
```

tools/archives/pyvo_integration/astronomical_archives.xml
bgruening: And we have it already :)


```
<param name="foo" type="text" value="foo 1">
    <option value="foo 1">Foo 1 Display</option>
    <option value="foo 2">Foo 2 Display</option>
</param>
```

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,1023 @@
...
+        <param name="basic_html_file" type="boolean" truevalue="true" falsevalue="false" label="Preview URLs as HTML"/>
```
volodymyrss, line 182:
```suggestion
        <param name="basic_html_file" type="boolean" truevalue="true" falsevalue="false" label="Return URL list as simple HTML"/>
```

tools/archives/pyvo_integration/astronomical_archives.xml
bgruening: Gosh Galaxy keeps surprising me.

tools/archives/pyvo_integration/astronomical_archives.xml
bgruening: A pyhon "None" I think.

tools/archives/pyvo_integration/astronomical_archives.xml
volodymyrss: Imagine how we feel :) Somehow almost every time I think of some feature of idea which could work here, I find that someone already tried it in Galaxy.

bgruening: haha, that sounds like a good toot! Made my day :)

tools/archives/pyvo_integration/astronomical_archives.xml
bgruening: Ok, it that case just make sure an output is actually selected. Not that by accident none of the outputs is selected. An alternative modeling of that UI would be a select field with optional=false (the default) and maybe as `display` checkboxes https://docs.galaxyproject.org/en/latest/dev/schema.html#id46

bgruening:
@francoismg @volodymyrss should I review again? When is your talk, we need to get this installed by then, do we?

volodymyrss:
> @francoismg @volodymyrss should I review again? When is your talk, we need to get this installed by then, do we?

Yes, it would be good. It's next Wednesday.
I made a couple of changes but we need @francoismg to finish what was being done.

volodymyrss:
When should we aim to install it so that it can appear on time? Assuming it might not workout at once.

bgruening:
Once its merged I can install it manually. So this can be fast. But I would say Monday latest?

tools/archives/pyvo_integration/astronomical_archives.xml
volodymyrss: Sounds good, this would be useful. @francoismg can you add it?

tools/archives/pyvo_integration/astronomical_archives.xml
bgruening: Yes it does. Since its a table field. I assume one ASCII is allowed, not whitespaces etc? If so we could restrict the input field here any in other places.

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,977 @@
...
+                <option value="optical">Optical</option>
+                <option value="uv">UV</option>
+                <option value="euv">EUV</option>
+                <option value="x-ray">X-ray</option>
+                <option value="gamma-ray">Gamma-ray</option>
+              </param>
+            </when>
+            <when value="archive">
+              <param name="archive" type="select" label="Astronomical archives">
+                <options from_data_table="astronomical_archives_1" />
```
bgruening, line 52: why has this a _1 ?

francoismg: my bad thanks, forgot to change after moving the code

bgruening:
Linting is failing because of Python flake8 errors.

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,977 @@
...
+          <filter>output_section['output_selection'] is not None</filter>
+        </data>
+        <data name="output_error" format="txt" label="${tool.name} -> Query Summary:"></data>
+    </outputs>
+    <tests>
+        <test expect_num_outputs="1">
+            <param name="output_selection" value="i"/>
+            <param name="number_of_files" value="1"/>
+            <conditional name="archive_selection">
+                <param name="archive_type" value="registry"/>
```
bgruening, line 164: it would be nice to have at least one test using the data_table.

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -0,0 +1,977 @@
...
+          <conditional name="query_selection">
+              <param name="query_type" value="obscore_query" />
+          </conditional>
+          <output name="output_error" count="1">
+            <assert_contents>
+                <has_text text="No obscore table in the archive"/>
+            </assert_contents>
+          </output>
+        </test>
+        <test expect_num_outputs="1">
```
bgruening, line 230: Mh, the tool does not fail when we have an error in the query?

Would that not be the expected behaviour from a users point of view.
You can test for error code and such. But this needs to be implemented in your python script I guess, that you really fail.

tools/archives/pyvo_integration/astronomical_archives.xml
francoismg: I had some when I was using "from file" instead of a data_table but when running planemo test I always got some "invalid value" or something like that.

I could try it again now that we moved to data_table

bgruening: maybe that one helps you https://github.com/galaxyproject/tools-iuc/tree/96f8a533278b4b6394aebd7a8f537513b0d29b1a/tools/recentrifuge

tools/archives/pyvo_integration/astronomical_archives.xml
francoismg: when a query fail we update the summary output to inform the user so he can possibly fix the error in the query, if the tool failed the user would not get the summary output right or am I mistaken?

bgruening: Yes, if it fails the outputs is not returned. But you get the stdout/stderr.

francoismg: ok I thought it would be more user friendly if we informed the user through some summary file but we could just output everything in the stdout if you think it would be better no problem

bgruening: Your call, I don't know :)
I'm happy to +1 it as it is :)

bgruening:
If you merge please create a pr to the tools repo I will make sure it gets installed during your night latest.
Thanks this is awesome!

volodymyrss:
@francoismg you want to add it? [Here](https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/594) is an example.

francoismg:
> @francoismg you want to add it? [Here](https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/594) is an example.

yea sure will try that

francoismg:
> @francoismg you want to add it? [Here](https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/594) is an example.

So just to be sure I understand correctly.

I should make a pr to the usegalaxy-eu-tools repo to update the astronomy yaml with the new tool reference and after that push the actual tool files to another repo is that correct?

You said you pushed your tools with planemo but that it caused some issues right? So after updating the yaml file I should make a pr to @bgruening  repo https://github.com/bgruening/galaxytools or the IUC repo https://github.com/galaxyproject/tools-iuc with the actual tool files, does that make sense?

volodymyrss:
> > @francoismg you want to add it? [Here](https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/594) is an example.
> 
> So just to be sure I understand correctly.
> 
> I should make a pr to the usegalaxy-eu-tools repo to update the astronomy yaml with the new tool reference and after that push the actual tool files to another repo is that correct?
> 
> You said you pushed your tools with planemo but that it caused some issues right? So after updating the yaml file I should make a pr to @bgruening repo https://github.com/bgruening/galaxytools or the IUC repo https://github.com/galaxyproject/tools-iuc with the actual tool files, does that make sense?

Not quite.

This tools-astro repository here has CI/CD which pushes to the toolsched. It [just failed](https://github.com/esg-epfl-apc/tools-astro/actions/runs/6072052691/job/16471817574) due to, apparently, [toolsched](https://toolshed.g2.bx.psu.edu/repository/browse_repositories_in_category?id=292389a45f1a238a) being down?.. It seems back now very slow to respond, let's see if CI completes.
                        
Once pushed to toolshed, the tools are added to the galaxy EU instance with https://github.com/esg-epfl-apc/usegalaxy-eu-tools/pull/1

francoismg:
> 

ok thanks, so just need to wait then

francoismg:
> 


Repository [suite_astronomicalarchivestools] does not exist in the targeted Tool Shed.
Failed to update repository 'suite_astronomicalarchivestools' as it does not exist on the main Tool Shed.

There's also that error, I followed the other yaml file you made and created another suite but apparently that might not be correct

volodymyrss:
The problem you quote was with access and I suspect only that, all the following problems follow it.

The problem now is this one:

```
<br/><b>astronomical_archives.xml<\/b> - This file refers to a file named <b>astronomical_archives_gen.loc<\/b>.  Upload a file named <b>astronomical_archives_gen.loc.sample<\/b> to the repository to correct this error.<br/>"}
```

francoismg:
> The problem you quote was with access and I suspect only that, all the following problems follow it.
> 
> The problem now is this one:
> 
> ```
> <br/><b>astronomical_archives.xml<\/b> 
> - This file refers to a file named <b>astronomical_archives_gen.loc<\/b>.  
> - Upload a file named <b>astronomical_archives_gen.loc.sample<\/b> to the repository to correct this error.<br/>"}
> ```

I pushed a commit with a loc.sample file but that doesn't seem to work since the pr has been merged (i think), will open another pr

PR #29, opened 2023-09-04
astroarchives: add sample

volodymyrss:


francoismg:
I don't think you have to include every row in the sample file at least from what I saw on the other tools, I think the purpose of the file is just to describe the data table

I will check once again but the few examples I looked at were this way

volodymyrss:
> I don't think you have to include every row in the sample file at least from what I saw on the other tools, I think the purpose of the file is just to describe the data table
> 
> I will check once again but the few examples I looked at were this way

Maybe, but it's a baseline list applicable for everybody, "sample" it is.

volodymyrss:
Since it is functionally the same as #30 I merge. Note that I needed to make a trivial change to the tool xml so that diff finds it and deploys. Hopefully this works.

francoismg:
> Since it is functionally the same as #30 I merge. Note that I needed to make a trivial change to the tool xml so that diff finds it and deploys. Hopefully this works.

ok great, let's see if that works now

volodymyrss:
Just FYI here is toolsched status https://uptime.statuscake.com/?TestID=YsdDV7MgNC it's not very up at all. I wonder if we missed some maintenance.

volodymyrss:
> See here, US seems to have problems: https://usegalaxy.org/

Oh well.
Although only TTS seems to be down. 

In TS I get [this](https://github.com/esg-epfl-apc/tools-astro/actions/runs/6073714898/job/16476883913):
```
You already own a repository named <b>suite_astronomicalarchivestools<\/b>, please choose a different name.", "err_code": 400008}
Repository [suite_astronomicalarchivestools] does not exist in the targeted Tool Shed
```

We had this problem before, it's something else.

but I think manual upload here worked:

https://toolshed.g2.bx.psu.edu/view/astroteam/astronomical_archives/0ddfc343f9f9

PR #30, opened 2023-09-04
added loc.sample file to archive tool data table

francoismg:


volodymyrss:
does it differ from #30  ?

francoismg:
> does it differ from #30 ?

what do you mean?

PR #40, opened 2023-10-20
Archives tool cone search

francoismg:
@volodymyrss

PR #41, opened 2023-10-20
fixed linting

francoismg:


tools/archives/pyvo_integration/astronomical_archives.py
```diff
@@ -695,7 +696,7 @@ def _set_cone_query(self):
         ts = 'target_selection'
         con = 'cone_object_name'
 
-        search_radius = self._json_parameters[qs][qse]['radius']
+        search_radius = self._json_parameters[qs][qsl]['radius']
```
bgruening, line 699: that looks more like a linting fix ... do you want to bump the version number?

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -682,7 +683,7 @@ MINIJ-PAS PDR201912 Catalogue (December, 2019) is based on scientific images in
 
 -----
 
-**ChiVO TAP** https://vo.chivo.cl/tap ChiVO TAP service
+**ChiVO TAP** https vo.chivo.cl/tap ChiVO TAP service
```
volodymyrss, line 686: why this?

francoismg: The shed_lint command check that every url returns a 200 and unfortunately those are having some problems right now, so I thought it was the best way to keep the information and still pass the checks. It can be reversed once the archive have fix their issues

tools/archives/pyvo_integration/astronomical_archives.py
francoismg: Not totally sure what you mean by that?

The version number has already been bumped in the previous pull request (https://github.com/esg-epfl-apc/tools-astro/pull/40) that adds a new feature, this one is only there to fix some linting issues related to the previous pr

tools/archives/pyvo_integration/astronomical_archives.xml
volodymyrss: ok, could you make a comment about that in the code so we know in the future?

tools/archives/pyvo_integration/astronomical_archives.py
volodymyrss: I guess @bgruening is pointing out that this change, `qse` => `qsl`, does not look like a linting change.
In fact, `qse` is not defined, so it was a typo, right @francoismg ?

Then, the question is if we need to update tool version. For this fix. Since the previous version did not actually get published to toolsched, it does not seem necessary to bump the version with this bug and lint fix, right @bgruening ?f

volodymyrss: I guess @bgruening is pointing out that this change, `qse` => `qsl`, does not look like a linting change.
In fact, `qse` is not defined, so it was a typo, right @francoismg ?

Then, the question is if we need to update tool version. For this fix. Since the previous version did not actually get published to toolsched, it does not seem necessary to bump the version with this bug and lint fix, right @bgruening ?f

bgruening: If its not published you don't need to bump the version, that is correct. I missed the fact that it was not pushed to the TS.

francoismg:
> > @volodymyrss merging is blocked until review
> 
> makes sense! If you are ready with the change, feel free to re-request review.

every checks pass, I retested it with planemo serve and test and everything seemed ok

volodymyrss (approved):

volodymyrss:
Ok it seems it is deployed https://toolshed.g2.bx.psu.edu/view/astroteam/astronomical_archives/667fc28d803c
Could you please make the PR for EU instance @francoismg ?

volodymyrss:
> Ok it seems it is deployed https://toolshed.g2.bx.psu.edu/view/astroteam/astronomical_archives/667fc28d803c Could you please make the PR for EU instance @francoismg ?

Could you please next time leave the reference here too so I am aware? 
Here it is for the record https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/635

francoismg:
> > Ok it seems it is deployed https://toolshed.g2.bx.psu.edu/view/astroteam/astronomical_archives/667fc28d803c Could you please make the PR for EU instance @francoismg ?
> 
> Could you please next time leave the reference here too so I am aware? Here it is for the record [usegalaxy-eu/usegalaxy-eu-tools#635](https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/635)

yes no problem, I had notified you on matrix but forgot to put the reference

PR #55, opened 2023-12-18
Live preview action

dsavchenko:


dsavchenko:
I don't have a permission to set reviewer in this repo, so please review @volodymyrss

volodymyrss (approved):
I think it's ok actually, also not too complicated. 
Maybe test-live should be added automatically by bot?

volodymyrss:
do you already have rights to merge @dsavchenko ? Anyway I merge, nice to have a look what it does. Let's add next changes in new PR as needed!

.github/workflows/live-preview.yaml
```diff
@@ -0,0 +1,75 @@
+name: Preview PR on galaxy.odahub.fr
+on:
+  pull_request:
```
volodymyrss, line 3: it should also run on change label @dsavchenko

dsavchenko: There is pull_request: labeled event there, isn't it enough?

PR #59, opened 2023-12-19
Update tool HESS to 0.0.1+galaxy0

dsavchenko:


tools/hess/hess_astro_tool.xml
```diff
@@ -0,0 +1,138 @@
...
+    <requirement type="package" version="3.8.2">matplotlib</requirement>
+    <requirement type="package" version="0.4.6">astroquery</requirement>
+    <requirement type="package" version="1.11.4">scipy</requirement>
+    <requirement type="package" version="1.2.3">oda-api</requirement>
+    <!--Requirements string 'nb2workflow[cwl,service,rdf,mmoda]>=1.3.30 
+' can't be converted automatically. Please add the galaxy/conda requirement manually or modify the requirements file!-->
+    <requirement type="package" version="7.13.0">nbconvert</requirement>
+    <requirement type="package" version="1.20.3">wget</requirement>
+  </requirements>
+  <command detect_errors="exit_code">ipython '$__tool_directory__/${_data_product._selector}.py'</command>
```
bgruening, line 15: why ipython?

tools/hess/hess_astro_tool.xml
```diff
@@ -0,0 +1,138 @@
...
+        <param name="Radius" type="float" value="2.5" label="Radius Units: deg" />
+        <param name="R_s" type="float" value="0.2" label="R_s Units: deg" />
+        <param name="Emin" type="float" value="100.0" label="Emin Units: GeV" />
+        <param name="Emax" type="float" value="10000.0" label="Emax Units: GeV" />
+        <param name="NTbins" type="integer" value="2" label="NTbins" />
+      </when>
+    </conditional>
+  </inputs>
+  <outputs>
+    <data label="${tool.name} -&gt; Image picture" name="out_Image_picture" format="auto" from_work_dir="picture_galaxy.output">
```
bgruening, line 64: I know format="auto" is the most simple solution and is good in the beginning. I just wanted to point out that is would be nice and also faster if we could specify a fixed value here.

tools/hess/hess_astro_tool.xml
```diff
@@ -0,0 +1,138 @@
...
+        <param name="R_s" value="0.2" />
+        <param name="Emin" value="100.0" />
+        <param name="Emax" value="10000.0" />
+        <param name="NTbins" value="2" />
+      </conditional>
+      <assert_stdout>
+        <has_text text="*** Job finished successfully ***" />
+      </assert_stdout>
+    </test>
+  </tests>
```
bgruening, line 137: A help section is missing. What about citations?

dsavchenko:
@bgruening this PR is automatically created by the bot https://github.com/oda-hub/oda-bot/pull/40 using nb2galaxy from https://github.com/oda-hub/nb2workflow/pull/102
It's based on the repo https://gitlab.renkulab.io/astronomy/mmoda/hess which we also use as the source for MMODA tools. 
I already activated this for some tools to see all stack in action as well as enable the preview of the PR on staging instance galaxy.odahub.fr
With this we will gradually approach to the state of tools which are ready for deployment on the toolshed. 

It's a bit of a work-in-progress, so some parts are missing. The generation of them is mostly implemented, but the corresponding pieces of info need to be added to the original repo by its author (@andriineronov in this case)

tools/hess/hess_astro_tool.xml
dsavchenko: As far as this is converted from notebook, which can contain some jupyter magic, using ipython to run the script assures they will be properly treated. It's not a great sacrifice to have an additional requirement, but the logic may be improved a bit to use it only when magic commands are actually there

tools/hess/hess_astro_tool.xml
dsavchenko: That's the best general way I was able to came up with. As far as tool description is generated from the notebook, and we can't be 100% sure of the outputs format without actually running it

bgruening:
@dsavchenko thanks for the explanations, appreciated!
Im aware of the autogeneration, I just wanted to give some comments in the hope the autogeneration could be improved - or ultimately the annotation of the original notebook.

PR #73, opened 2024-01-25
Update tool MWA to 0.0.4+galaxy0

dsavchenko:


tools/mwa/mwa_astro_tool.xml
```diff
@@ -0,0 +1,117 @@
...
+			volume = {464},
+			number = {1},
+			pages = {1146-1167},
+			doi = {10.1093/mnras/stw2337},
+			archivePrefix = {arXiv},
+			eprint = {1610.08318},
+			primaryClass = {astro-ph.GA},
+			adsurl = {https://ui.adsabs.harvard.edu/abs/2017MNRAS.464.1146H},
+			adsnote = {Provided by the SAO/NASA Astrophysics Data System}
+		}</citation>
```
bgruening, line 115:
```suggestion
    <citation type="doi">10.1093/mnras/stw2337</citation>
```

tools/mwa/mwa_astro_tool.xml
```diff
@@ -0,0 +1,117 @@
...
+          <option value="GLEAM 103-134 MHz">GLEAM 103-134 MHz</option>
+          <option value="GLEAM 139-170 MHz">GLEAM 139-170 MHz</option>
+          <option value="GLEAM 170-231 MHz" selected="true">GLEAM 170-231 MHz</option>
+          <option value="GLEAM 72-103 MHz">GLEAM 72-103 MHz</option>
+        </param>
+      </when>
+      <when value="Spectrum">
+        <param name="src_name" type="text" value="1ES 0229+200" label="src_name" />
+        <param name="RA" type="float" value="38.202562" label="RA Units: deg" />
+        <param name="DEC" type="float" value="20.288191" label="DEC Units: deg" />
```
bgruening, line 43: just a very silly question ... is `deg` degree? And do you think it would make sense to have a type="deg"? What would such a type have as properties?

You could also add a min="" and max="" here I think.

tools/mwa/mwa_astro_tool.xml
dsavchenko: I imagined this is a preferred way. The bot is just currently only doing dummy copying. This will improve.

tools/mwa/mwa_astro_tool.xml
dsavchenko: What we are currently dealing with, out of galaxy context, is the structured input type, e.g. a table that gets encoded as a dict of lists. At some point, we will need to find a way to have such a form of input in galaxy. So we will most likely try to extend this part

PR #78, opened 2024-01-31
User defined archive access url (local branch PR)

### LINKED ISSUE #38: add free field for custom url
in "Archive Selection" we have:

* single archive
* query all

to add:

* custom archive url

francoismg:
Add the possibility for users to use a custom access url in order to query not supported archives


dsavchenko:
Preview is available https://galaxy.odahub.fr/?tool_id=astronomical_archives_pr78&version=latest

dsavchenko:
I fixed this https://github.com/esg-epfl-apc/tools-astro/actions/runs/7724870060/job/21057927546?pr=78
But do we really need such a strict linting?

francoismg:
> I fixed this https://github.com/esg-epfl-apc/tools-astro/actions/runs/7724870060/job/21057927546?pr=78 But do we really need such a strict linting?

Ok thanks, I was waiting for feedback to see if there was some changes to do before fixing the linting.

And yes linting is way too strict (more strict than what they use for galaxy core too btw) but i'm not sure we can change it ourselves since it's used to validate things for the toolshed

dsavchenko:
[This failure](https://github.com/esg-epfl-apc/tools-astro/actions/runs/7744736561/job/21119159360?pr=78) is very cryptic. But doesn't this entire workflow duplicate other ones?

francoismg:
> [This one](https://github.com/esg-epfl-apc/tools-astro/actions/runs/7744736963/job/21119353698?pr=78) is also a bit cryptic. But probably you understand what's the reason, @francois ?

Yes it's because an url from the tool documentation is unavailable at the moment, I will just format it so it doesn't appear like a url during parsing and it will pass the test.

Maybe that will fix the first error too cause if I remember correctly the first workflow success depend on the success of the following ones.

francoismg:
> [This one](https://github.com/esg-epfl-apc/tools-astro/actions/runs/7744736963/job/21119353698?pr=78) is also a bit cryptic. But probably you understand what's the reason, @francois ?

The second one is fixed now, but the first one still fails I will have to check it a bit

francoismg:
The tests are all good when running "planemo lint" locally but the workflow is a bit different here

I saw that in the log too, is it possible that the other tool in the repo are tested alongside this one and causing the error?

linting tools/astropytools/fits2bitmap.xml
Linting tool /home/runner/work/tools-astro/tools-astro/tools/astropytools/fits2bitmap.xml
.. CHECK: Tool defines a name [astropy fits2bitmap].
.. CHECK: Tool defines an id [astropy_fits2bitmap].

francoismg:
Ok The same error is appearing in every pull request so I think that the problem is not linked to that tool or PR

Every recent PR has the same errors : 

File "src/lxml/xmlschema.pxi", line 90, in lxml.etree.XMLSchema.__init__
lxml.etree.XMLSchemaParseError: complex type 'Output': The content model is not determinist., line 5162

and the bits about fits2bitmap

@dsavchenko

francoismg:
I tried to modify the worflow to run it only on some tools and it seems that you get the same error whatever tools you test, also I saw that there's the same error in commits from weeks ago so it looks like the problem is not recent

francoismg:
I was able to fix the error by updating planemo version in environment.yml will check to be sure and then make a PR

volodymyrss:
> I was able to fix the error by updating planemo version in environment.yml will check to be sure and then make a PR

Cool, thanks! I can also see it on [preview](https://galaxy.odahub.fr/root?job_id=6e4500518907a33e&__vkey__=1706820308728).

Is it possible to restrict text field pattern? So it only allows URLs..

volodymyrss:
After looking around in the preview (I really find it useful, much faster than pull, planemo s, etc) I confirm it seems to do what we want. Thanks @francoismg !

image

volodymyrss (approved):

dsavchenko:
Do we need to update the version of the tool manually? Or it will be bumped automatically?

francoismg:
> > I was able to fix the error by updating planemo version in environment.yml will check to be sure and then make a PR
> 
> Cool, thanks! I can also see it on [preview](https://galaxy.odahub.fr/root?job_id=6e4500518907a33e&__vkey__=1706820308728).
> 
> Is it possible to restrict text field pattern? So it only allows URLs..

I don't think there's an "url" input type or an other way to enforce a specific string format but I will check if I found something. There's still an url validation in the python part.

francoismg:
> Do we need to update the version of the tool manually? Or it will be bumped automatically (I didn't find it mentioned) ?

Yes we do as far as I know, at least that's what I did for the previous versions

volodymyrss:
> > Do we need to update the version of the tool manually? Or it will be bumped automatically (I didn't find it mentioned) ?
> 
> Yes we do as far as I know, at least that's what I did for the previous versions

Was @dsavchenko case of version updates for tools relevant here? Maybe there is a straighforward way to bump?

dsavchenko:
> Is it possible to restrict text field pattern? So it only allows URLs..

I see this https://docs.galaxyproject.org/en/latest/dev/schema.html#validators-for-textual-inputs-text-select
I'm not sure if it executes client-side or server-side upon submission. 
Anyway, it would be good to set this to validate with the same regex you use inside the script.

francoismg:
> > Is it possible to restrict text field pattern? So it only allows URLs..
> 
> I see this https://docs.galaxyproject.org/en/latest/dev/schema.html#validators-for-textual-inputs-text-select I'm not sure if it executes client-side or server-side upon submission. Anyway, it would be good to set this to validate with the same regex you use inside the script.

Thanks seems good I will try to add it, apparently it stops tool submission so it will behave like a frontend validation I think

dsavchenko:
> Was @dsavchenko case of version updates for tools relevant here? Maybe there is a straighforward way to bump?

In automatic tool updates, the bot just increments the patch version (like 0.1.2+galaxy0 --> 0.1.3+galaxy0) upon creating a PR. That's not at all obvious how to automatically determine which version part needs to be bumped. We can think about some CI action that modifies the version based on, e.g. label, so that manual intervention is simple. 


In the context of this PR, you just need to increase version manually in the xml. As this is kind of new feature, I suggest increasing the minor version.

volodymyrss:
> > Was @dsavchenko case of version updates for tools relevant here? Maybe there is a straighforward way to bump?
> 
> In automatic tool updates, the bot just increments the patch version (like 0.1.2+galaxy0 --> 0.1.3+galaxy0) upon creating a PR. That's not at all obvious how to automatically determine which version part needs to be bumped. We can think about some CI action that modifies the version based on, e.g. label, so that manual intervention is simple.

It can be based on labels. It would fit in the github flow for galaxy tools. But it's not the highest priority https://github.com/esg-epfl-apc/tools-astro/issues/85

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="astronomical_archives" name="Astronomical Archives (IVOA)" version="0.9.1">
+<tool id="astronomical_archives" name="Astronomical Archives (IVOA)" version="0.9.2">
```
dsavchenko, line 1: Probably 0.10.0 ? This PR introduces a new feature

volodymyrss (changes requested):
please add a regex for url validation

volodymyrss:
@francoismg says the validator does not work on Galaxy. To be figured out.

francoismg:
> @francoismg says the validator does not work on Galaxy. To be figured out. Simple regex works, but URL regex does not validate and does nothing. @francoismg will update here with details.

That's what I got, no message are displayed in the interface even when post request returns an error (same error whether url is valid or not)

I had the same behavior one or two times with a simple regex but it was working correctly most of the time

I pasted the regex in a standalone python script and was working fine

Maybe there's some specific syntax in the regex that's not processed correctly or something like that

Screenshot 2024-02-13 at 16 30 12

dsavchenko:
In your script, you use the `IGNORECASE` flag, but here it's not possible to add flags, I believe
https://EXAMPLE.COM/FOOBAR will be accepted, you can check

So we need to explicitly allow lowercase letters. 
Also, what's the point of using non-capturing groups? Probably, the regex will be a bit more readable without these `?:`

dsavchenko:
> Regarding ?, I just used the regex Django used to validate url cause it seemed more reliable to use something that has already been vetted but I can remove them if you think it will be more readable

This changes nothing unless you refer to groups afterwards. So not really critical.

volodymyrss:
Does not seem to work with https://datalab.noirlab.edu/tap

francoismg:
> Does not seem to work with https://datalab.noirlab.edu/tap

I guess you have tried with the all caps version like denys suggested.

If that's the case I think the problem is that noirlab takes really long to respond to the first query when pyvo tries to reach url/tables. When I tried it outside the tool some times ago it took more than 10 seconds if I remember correctly and the timeout is set to 10 so that's probably what's happening. I could try to lift the timeout for user provided url and since there's only one url it won't be as much a problem as with keyword search.

volodymyrss:
> > Does not seem to work with https://datalab.noirlab.edu/tap
> 
> I guess you have tried with the all caps version like denys suggested.
> 
> If that's the case I think the problem is that noirlab takes really long to respond to the first query where pyvo tries to reach url/tables. When I tried it outside the tool some times ago it took more than 10 seconds if I remember correctly and the timeout is set to 10 so that's probably what's happening. I could try to lift the timeout for user provided url and since there's only one url it won't be as much a problem as with keyword search.

Please try to make sure it works for this URL?

Also maybe add this URL to the tool test.

tools/archives/pyvo_integration/astronomical_archives.xml
```diff
@@ -47,6 +48,11 @@
                 <options from_data_table="astronomical_archives" />
               </param>
             </when>
+            <when value="custom">
+              <param name="access_url" type="text" label="TAP archive access url">
+                <validator type="regex" message="Url seems invalid">^(?:http)s?://(?:(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+(?:[A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?)|)(?::\d+)?(?:/?|[/?]\S+)$</validator>
```
volodymyrss, line 53:
```suggestion
                <validator type="regex" message="URL seems invalid">^(?:http)s?://(?:(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+(?:[a-zA-Z]{2,6}\.?|[a-zA-Z0-9-]{2,}\.?))(?::\d+)?(?:/?|[/?]\S+)$</validator>
```

francoismg:
> > > Does not seem to work with https://datalab.noirlab.edu/tap
> > 
> > 
> > I guess you have tried with the all caps version like denys suggested.
> > If that's the case I think the problem is that noirlab takes really long to respond to the first query where pyvo tries to reach url/tables. When I tried it outside the tool some times ago it took more than 10 seconds if I remember correctly and the timeout is set to 10 so that's probably what's happening. I could try to lift the timeout for user provided url and since there's only one url it won't be as much a problem as with keyword search.
> 
> Please try to make sure it works for this URL?
> 
> Also maybe add this URL to the tool test.

Yes no problem will do that and will add a new test case for the new feature with that specific url

volodymyrss:
And let's also add vo paris hess he to the test.

volodymyrss:
How is it going here?

francoismg:
> How is it going here?

Still have to add the two new test cases for vo hess paris and noir lab.

Also need to remove the capturing group from the regex.

Will do that after the meeting and besides that I think we were good if you don't have any more requests.

I will push today and request a new review

volodymyrss:
@francoismg mention in an email that there was an issue querying noirlab tab server. 

Previously @burnout87 implemented in [astroquery for this tap server](https://github.com/astropy/astroquery/pull/2210) so maybe you can please advise one @francoismg pushes the problematic code.

francoismg:
> > Previously @burnout87 implemented in [astroquery for this tap server](https://github.com/astropy/astroquery/pull/2210) so maybe you can please advise one @francoismg pushes the problematic code.
> 
> Here @francoismg can check the code I wrote some time ago to build a TAP query
> 
> https://github.com/oda-hub/astroquery/blob/7cc8d8b12829a7d3696266f333da1cfca3fe3d01/astroquery/desi/core.py#L21-L54
> 
> Please ask any question you might have, so I can help you with the implementation here

Thanks for the help, problem was that the tap schema is not up to date and was crashing when trying to fetch a schema that doesn't exist anymore for a specific table and I was also trying to query a table whose schema was available but doesn't exist anymore in the db.

It's all working now and I have updated the tests accordingly, I also added a list of slow archives (only noir lab right now) so the timeout decorator switch dynamically to a higher value when trying to access some useful but slow archives.

Regex for url validation has also been modified regarding previous comments.

Will check that everything is working fine and that tests are executed correctly and will push for review.

francoismg:
Still one bug to fix when writing noir lab user chosen url field to the csv file

francoismg:
> Still one bug to fix when writing noir lab user chosen url field to the csv file

write bug is fixed, also there's seems to be some differences between regex processing from the xml and with python so the regex are now a bit different but are working correctly

volodymyrss:
Actions broken for unknown reason

dsavchenko:
I already raised this question. There are always problems with this specific action. And it seems to have just a duplicated functionality with the other more elaborated ones. 
I suspect, looking at the history, this one was created manually as a first try, but then "official" CI/CD was added to the repo. @volodymyrss please confirm
So we are safe to just remove it completely, I think

volodymyrss:
> I already raised this question. There are always problems with this specific action. And it seems to have just a duplicated functionality with the other more elaborated ones. I suspect, looking at the history, this one was created manually as a first try, but then "official" CI/CD was added to the repo. @volodymyrss please confirm So we are safe to just remove it completely, I think

No idea, I took an example provided, feel free to remove in this PR if it is indeed duplicated.

dsavchenko:
@francoismg [this tool linting fail](https://github.com/esg-epfl-apc/tools-astro/actions/runs/8345873321/job/22842044506?pr=78) is probably again about unresolvable link in description, I'm not sure, could you look on it?

francoismg:
> @francoismg [this tool linting fail](https://github.com/esg-epfl-apc/tools-astro/actions/runs/8345873321/job/22842044506?pr=78) is probably again about unresolvable link in description, I'm not sure, could you look on it?
> 
> But tests pass!

yes looks like a documentation url is not reachable, I will break it so it is not parsed as a url

Also seems that there's one line too long and some whitespaces somewhere, will fix that too

dsavchenko:
@francoismg any progress on fixing linting?

francoismg:
> @francoismg any progress on fixing linting?

yes the problem is that the regex triggers some errors cause it's treated like a regular string so I will have to replace the regex with some urllib functions and parse the url myself like we mentionned earlier so it won't be the same validation in the xml and the python script. I was working on the jsvis right now but I will make and push the archive tool changes today

francoismg:
> > @francoismg any progress on fixing linting?
> 
> yes the problem is that the regex triggers some errors cause it's treated like a regular string so I will have to replace the regex with some urllib functions and parse the url myself like we mentionned earlier so it won't be the same validation in the xml and the python script. I was working on the jsvis right now but I will make and push the archive tool changes today

or maybe use the raw string thing r'' maybe it will make flake8 understand that it's not escaping blackslash and we could keep the regex, I will try that before switching to urllib

dsavchenko:
> or maybe use the raw string thing r''

Even regardless flake8, my experience is that it's always better to use raw string for the regex

francoismg:
> > or maybe use the raw string thing r''
> 
> Even regardless flake8, my experience is that it's always better to use raw string for the regex

ok will do that then thanks, it was like that before but I removed it for some reason, I should have kept it

tools/archives/pyvo_integration/astronomical_archives.py
```diff
@@ -1250,9 +1280,9 @@ def write_urls_to_output(urls: [], output, access_url="access_url"):
         with open(output, "w") as file_output:
             for url in urls:
                 try:
-                    file_output.write(url[access_url] + ',')
+                    file_output.write(str(url[access_url]) + ',')
                 except Exception:
-                    error_message = "url field not found for url"
+                    error_message = "url field "+access_url+" not found for url"
```
dsavchenko, line 1285:
```suggestion
                    error_message = f"url field {access_url} not found for url"
```

tools/archives/pyvo_integration/astronomical_archives.xml
dsavchenko:
```suggestion
<tool id="astronomical_archives" name="Astronomical Archives (IVOA)" version="0.10.0">
```

dsavchenko:
And for some "line too long" I would've just used `#noqa` because it really seems too restrictive. Sometimes, following linter suggestions only worsen code readability

francoismg:
> And for some "line too long" I would've just used `#noqa` because it really seems too restrictive. Sometimes, following linter suggestions only worsen code readability

Yes it's way too restrictive I have too keep one working version where I dont follow every suggestions cause the linted version is so unpractical

I thought we had to follow every rules to adapat to galaxy standards since it's pushed to the tool shed but ok if we can just make it ignore some things it's great I will add that too where it makes sense

tools/archives/pyvo_integration/astronomical_archives.py
```diff
@@ -752,6 +777,11 @@ def run(self):
 
             for archive in self._archives:
                 try:
+
+                    if archive.access_url in ARCHIVES_TIMEOUT_BYPASS:
+                        archive.get_resources = \
+                            timeout(40)(TapArchive.get_resources.__get__(archive))
+
```
dsavchenko, line 784:
```suggestion
                            timeout(40)(TapArchive.get_resources.__get__(archive))  # noqa: E501

```

tools/archives/pyvo_integration/astronomical_archives.py
```diff
@@ -1305,6 +1335,12 @@ def collect_resource_keys(urls_data: list) -> list:
                     resource_keys.append(key)
         return resource_keys
 
+    @staticmethod
+    def is_valid_url(url: str) -> bool:
+        regex_url = re.compile('^https?://(?:[A-Za-z0-9-]+\.)+[A-Za-z]{2,6}(?::\d+)?(?:/[^\s]*)?$')
+
```
dsavchenko, line 1341:
```suggestion

```

dsavchenko (approved):

francoismg:
> @volodymyrss I made final polishing so that we have green mark here. I think we can merge right after your review. @francoismg please protest if you wanted to add something to this PR

No it's all good I was just going to try to add the raw string for the regex and try the noqa flag but since you did it already it's fine with me thanks.

volodymyrss (approved):
Looks good I think, I also tried it with other URLs I had. Thanks!

PR #79, opened 2024-01-31
Update tool CRbeam to 0.0.1+galaxy0

dsavchenko:


volodymyrss:
Last update failed, @dsavchenko .

dsavchenko:
@bgruening could you help? For the tools that are not yet in the toolshed we get
```
bioblend ERROR: GET: error 404: b'{"status": "error", "message": "No repository named crbeam_astro_tool found with owner astroteam"}', 0 attempts left
Failed linting
```
in the linting action. 
It wasn't like this before.

dsavchenko:
> The other `WARNING (TestsCaseValidation)` is also something new that I don't understand

This will be fixed with https://github.com/oda-hub/nb2workflow/pull/247 
The issue was supposedly resolved by https://github.com/galaxyproject/galaxy/pull/19790, but we still see it here. Anyway, we don't require parameters to start with underscore in the generated tools

dsavchenko:
@okolo I polished this, and now it passes all checks (I defined a bit more permissive linting that will apply to other tools as well once this is merged). 
Please inspect my changes in the parent renku repo. I tried to be conservative, but just in case  
I think it's ready to be published. I ask for review of @volodymyrss

okolo:
@dsavchenko thank you very much! I am fine with the changes you have introduced in CRbeam repo. I also see that Galaxy tool pipeline failed on renku. Is it expected?

dsavchenko:
> I also see that Galaxy tool pipeline failed on renku. Is it expected?

Kind of expected. I migrated a galaxy bot to the prod cluster while staging is unavailable and there were some problems, so the bot set the failed status and then hasn't been triggered again

volodymyrss (approved):

PR #80, opened 2024-01-31
Update tool SGWB to 0.0.1+galaxy0

dsavchenko:


volodymyrss:
Why is it still failing, this we saw before I think?

dsavchenko:
> Why is it still failing, this we saw before I think?

The bot was stuck, so the change from the last commit hadn't been propagated. I also changed the repo config to enable "merge main branch". One problematic action is removed in main already. Now all CI/CD will pass, I suppose.

volodymyrss (approved):
I added the review just to satisfy formal condition.
We should really use a bot, and then @dsavchenko can review too.
We don't have a bot in this org yet I guess. Let's just add special one, @dsavchenko ? If you have trouble with it, I will do.

dsavchenko:
> We don't have a bot in this org yet I guess. Let's just add special one, @dsavchenko ? If you have trouble with it, I will do.

OK, I will deal with it. I thought of the one you recently created in the other org, but separating is probably better

volodymyrss:
> > We don't have a bot in this org yet I guess. Let's just add special one, @dsavchenko ? If you have trouble with it, I will do.
> 
> OK, I will deal with it. I thought of the one you recently created in the other org, 

> but separating is probably better

Exactly. For security and easier management of different functions.

PR #81, opened 2024-01-31
Update tool MWA to 0.0.1+galaxy0

dsavchenko:


volodymyrss:
@dsavchenko , do you know why this one is not updating and also failing like that:

image

dsavchenko:
It's updated. But what's going on with matplotlib dependency, that I don't understand for a moment

dsavchenko:
Other weirdness is why karabo script is in the PR...

dsavchenko:
> Other weirdness is why karabo script is in the PR...

Here is a fix https://github.com/oda-hub/oda-bot/pull/68

It happened because karabo uses non-standard conda channels, so that converter can't convert it and produces exception. But a script is already generated at that point. In git it's untracked, so stays there while changing branch

dsavchenko:
> @dsavchenko , do you know why this one is not updating and also failing like that:

The problem was with instance config, namely having both `conda-forge,defaults` in channels leaded to libmamba exception. And galaxy silently gives up installing requirement in this case, so the environment didn't contain needed packages. 

It works now

PR #83, opened 2024-01-31
Update tool GRB detection to 0.0.1+galaxy0

dsavchenko:


dsavchenko:
@volodymyrss this one is used in https://workflowhub.eu/workflows/766 which we need to finalize.
It lacks Galaxy help and citation. Also, only one notebook out of two is needed in the workflow and ready to be used in Galaxy (the other one has pypi-only requirements).
As we no more can create new repos in Renku, two ways to go would be manual editing of the tool definition here before publishing or temporarily masking the second notebook (which is way easier, but has side effect on MMODA, of course).

volodymyrss:
Its unlikely even of not impossible I will have time. Could you just add something sensible?

PR #91, opened 2024-02-19
Update tool lightcurve-analysis to 0.0.1+galaxy0

dsavchenko:


volodymyrss:
With first commit the default was used instead of input.

With second commit the tool disappeared :(

https://galaxy.odahub.fr/

dsavchenko:
> With second commit the tool disappeared :(

This is some transient problem which I already saw before. The last installed tool may randomly disappear. 

Probably, Galaxy requires restarting upon adding the tool (at least, this seemed to help). Currently, I'm injecting tools to the live instance. 
I will add the restart to the playbook. It may introduce some minor downtime, but it's not critical here. We can explore how tool registering is happening in Galaxy in more details in the future.

dsavchenko:
> With first commit the default was used instead of input.

This is still the case. I'm having a look

volodymyrss:
> > With second commit the tool disappeared :(
> 
> This is some transient problem which I already saw before. The last installed tool may randomly disappear.
> 
> Probably, Galaxy requires restarting upon adding the tool (at least, this seemed to help). Currently, I'm injecting tools to the live instance. I will add the restart to the playbook. It may introduce some minor downtime, but it's not critical here. We can explore how tool registering is happening in Galaxy in more details in the future.

Ok thanks, the tool is back.

The tool fails somehow, looks like the parameter is not substituted https://galaxy.odahub.fr/datasets/98e8737e708a5bb0/details

dsavchenko:
> The tool fails somehow, looks like the parameter is not substituted

Fixed

volodymyrss:
This works pretty well!

https://galaxy.odahub.fr/u/volodymyr/w/grb-analysis-workflow

Here I normalize and then combine HESS and SPI-ACS data. It's a of a mock-up of course, I should have started from image or spectra. Or maybe I refine the LC if I find some common variable source, probably the BOAT GRB.

image

dsavchenko:
@volodymyrss 
This passes tests but misses help and citations. Do you have time this week to finalize? 
It's used in https://workflowhub.eu/workflows/766, and it would be great to polish the tools so that we can mark the workflow stable, and it will be able to run on usegalaxy.eu (the grb detection tool may be more difficult though, I will comment there)

dsavchenko:
After looking into notebooks, this seems to be really a mock-up version of the tool. 

The mode is only used in the "detect" notebook and the only value that matters is "Background". 
The peak detection is a simple max, does it make much sense? 
The "correlate" one does nothing but plotting both lightcurves (and no peak detection there at all).

I'm not sure, I can quickly improve the tool, not being very skilled in time-domain. Should we leave the workflow as WIP @volodymyrss ?

volodymyrss:
Lets leave it maybe. Ok could you remind me Friday if I do not do it?

volodymyrss:
Lets leave it maybe. Ok could you remind me Friday if I do not do it?

PR #96, opened 2024-02-26
Update tool IceCube to 0.0.1+galaxy0

dsavchenko:


tools/icecube/icecube_astro_tool.xml
```diff
@@ -0,0 +1,108 @@
...
+    <requirement type="package" version="6.0.0">astropy</requirement>
+    <requirement type="package" version="1.26.4">numpy</requirement>
+    <requirement type="package" version="1.12.0">scipy</requirement>
+    <requirement type="package" version="2.25.2">iminuit</requirement>
+    <requirement type="package" version="3.8.3">matplotlib</requirement>
+    <requirement type="package" version="1.20.3">wget</requirement>
+    <requirement type="package" version="0.70.16">multiprocess</requirement>
+    <requirement type="package" version="1.2.12">oda-api</requirement>
+    <requirement type="package" version="7.16.1">nbconvert</requirement>
+    <!--Requirements string 'skyllh
```
volodymyrss, line 13: why can it not be, @dsavchenko ?

dsavchenko: Seems it's not in conda
So currently no way to create the tool this way...

volodymyrss: As a workaround, can you just do pip install while running the tool?..

dsavchenko: It will work in our staging installation, I think. Might not be fully universal with other job runners.

But this is a really weird way, not sure, we want to include it in the generator

volodymyrss: On the other hand, I am not sure why galaxy does not allow pip. renkulab for example [allows this](https://github.com/SwissDataScienceCenter/renku-project-template/blob/ee4cf00c52c6cd45363a40b55540343e9ecf70af/python-minimal/Dockerfile#L45-L50) and it works fine for a large number of projects.
@bgruening , I am sure you saw this question before, and maybe even we asked, but is it really not possible to use pip packages in galaxy?

Anyway, since we auto-generate tool here, we can consider this extra pip install as part of bootstrapping, the user never sees it directly.

> It will work in our staging installation, I think. Might not be fully universal with other job runners.

Why not? Because there is no writable directory? Could be even --user assuming there is a writable home. Or it can be created.

> But this is a really weird way, not sure, we want to include it in the generator

It could be a fallback, allowing to show what capacity is provided by adding the tool to galaxy. This can motivate the package authors to put to conda. We probably do not want to push it to toolsched. Maybe the PR can be labeled as so. PRs with incomplete dependencies should be distingiushed.

bgruening: Sorry, only saw this no. 
dependency managers are just another plugin layer in Galaxy. brew, conda, modules ... ofc you can add pip if you like. The question we need to ask is where to stop, do we now need to support all language-specific package managers? How good are they? We still do want to move to containers, isn't it? How do we ensure containers are available for all the packages?

On the other side, adding pypi packages to conda-forge is so trivial nowadays and the entire OS community benefits from it.

Issue #64, opened 2024-01-15
add github action which updates IVOA tap server list


volodymyrss:

francoismg:
seems to be working fine https://github.com/esg-epfl-apc/tools-astro/pull/137

PR #107, opened 2024-04-09
Github action for updating archives list

francoismg:
As discussed in https://github.com/esg-epfl-apc/tools-astro/issues/64


francoismg:
@volodymyrss @dsavchenko

.github/workflows/update-archives-list.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: run archives script
+        run: python fetch_archives.py
+
+      - run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git add .
+          git commit -m "updated archives file"
+          git push
```
dsavchenko, line 34: Will the new revision of the tool be published to the toolshed if the action pushes directly to master?
PR could be better option anyway.

Also, I'm not sure, but as long as we change the integral part of the tool here, we need to bump its version, right?

fetch_archives.py
```diff
@@ -0,0 +1,30 @@
+import pyvo
+
+
+def fetch_archives():
+    archive_list = pyvo.registry.search(servicetype="tap")
+
+    with open('tools/archives/pyvo_integration/tool-data/astronomical_archives_gen.loc', 'w') as f, open('tools/archives/pyvo_integration/tool-data/astronomical_archives_gen.loc.sample', 'w') as f1:
```
dsavchenko, line 7: Could you remind me why do we need both files? They look completely the same

dsavchenko (changes requested):
There often occurs the linting problem with unavailable urls in tool description (help section, I suppose). Could we also deal with it in this action?

fetch_archives.py
francoismg: If I remember correctly you need to have a sample file otherwise it will generate some error when trying to publish the tool (we had that problem when we tried the first time)

The first version of the sample file only had one line of data but then we just used the same content for simplicity sake I guess. Will try to find the pull request where we talked about that and post it there

.github/workflows/update-archives-list.yml
volodymyrss: > The action could parse the help section and make a call for every urls and rewrite them if they are unavailable yes, I could make another script that does that and run it in the action or make another action specifically for that. Since it will modify the tool xml file maybe it would be better to have a specific action for that
> 
> If we don't want automatic push to toolshed when pushes are automatised we could discriminate against specific user inside the toolshed action right? Or yes it could be a PR rather than direct push and then it will be manually merged. Is that what you meant?

I also think PR is better at this time, although it may become a bit annoying to merge PRs every week, it's actually good to know what's being added at least at first.

> Regarding the version since it's only data files I don't think it would warrant a version bump but I'm not sure what that implies toolshed and usegalaxy wise.

I suppose tool data is part of the tool, so version change is warranted. But TBD.

> 
> I'm not sure of what happens if you push to toolshed without changing version, will the targeted version be overwritten?
> 
> Regarding galaxy.eu I think they pull the code directly from the toolshed when you access the site (I don't think they have all the tools locally) so if the usegalaxy tool version list already has the targeted version and the toolshed overwrite the code when changes are pushed to the same version it should be fine. But I'm really not sure about all that I will check if I can find more info

Is it possible that it pulls the version

volodymyrss:
@francoismg do we have something to finish and conversations to resolve here? Could you please have a look?

volodymyrss (commented):
takes 10 min to complete and a bit more to check.

.github/workflows/update-archives-list.yml
volodymyrss: We agreed to increase the version.

francoismg:
Actions need to be granted the right to create pull request in settings -> actions

volodymyrss:
> Actions need to be granted the right to create pull request in settings -> actions

Ok, done. Could you please check now?

francoismg:
> > Actions need to be granted the right to create pull request in settings -> actions
> 
> Ok, done. Could you please check now?

seems all good https://github.com/esg-epfl-apc/tools-astro/pull/111

I should just make sure that it will only contains the commits with the list update otherwise it will always try to push the new action to main

volodymyrss:
let's do it Thursday night once a week

volodymyrss (approved):

volodymyrss:
@dsavchenko could you please review?

fetch_archives.py
dsavchenko: Ok, the sample is needed when using "data managers", that are used to update *.loc files (normally, these files are used to keep info about "reference data"). As soon as there is no such "reference datasets" used by the tool, we stick to direct update. 
Sample file can be empty, only having column description. But if it works like this, let's not touch it:)

.github/workflows/update-archives-list.yml
dsavchenko: The question is still a case. The version needs to bumped in tool xml upon creating the PR.

dsavchenko (changes requested):
The version of the tool needs to be updated here (discussion above)

francoismg:
> The version of the tool needs to be updated here (discussion above)

Tool version update is working on my local test setup, just need to rework fetch_archives script so the sample file only contains columns description and first row and then will test it against the real repo

volodymyrss:
Thanks for the update, @francoismg , could you push your WIP here?

francoismg:
> Thanks for the update, @francoismg , could you push your WIP here?

WIP has been updated to work with pull request and is pushed, should work but had a pyvo exception last few runs so couldn't validate everything will try again in a moment

francoismg:
Everything is working on my side but I can't test it directly here cause workflow dispatch is not available (not sure if permission issue or something else), we can try to merge and see if it's working if it's ok with you @volodymyrss

volodymyrss (approved):

dsavchenko (approved):

PR #125, opened 2024-09-11
fix conda version

bgruening:


bgruening:
This version never existed, no idea why the test has passed. But I saw a container problem here: https://github.com/BioContainers/multi-package-containers/pull/3195

dsavchenko:
I believe this tool is a prototype of a more general oda request 
But it can be viewed as a fully functional tool for SPI-ACS request on the other hand. Should we finalise @volodymyrss ? What's the status of SPI-ACS service ?

PR #178, opened 2025-02-07
Update ci/cd

dsavchenko:
Following current versions in https://github.com/galaxyproject/tools-devteam
Also, remove R-related actions


dsavchenko:
I will merge now to see it in action

PR #205, opened 2025-04-23
Fix automatic deployment and update astropy tools versions

dsavchenko:


volodymyrss (approved):

dsavchenko:
@Andrei-EPFL @volodymyrss updated astropy tools are available both on our staging and usegalaxy.eu https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/853

PR #212, opened 2025-05-02
Bypassing planemo url linting

francoismg:
Possible fix for this https://github.com/esg-epfl-apc/tools-astro/pull/208#issuecomment-2835512860

I made some test locally but it seems that with latest versions of planemo the --urls flag needs to be added to lint urls otherwise the check is skipped

Also urls is not skippable using planemo lint --skip 'urls' because it's not part of the list of linting tests

I'm not totally sure "tool_urls" is responsible for that as I couldn't find it in the documentation but looking at this issue https://github.com/galaxyproject/planemo/issues/677 it seems that at some point it was part of the lint test cases so I guess it's worth a shot 

Possibly due to the use of cache action the ci is still using some old version of planemo where urls are checked by default but not sure about that

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -103,7 +103,7 @@ jobs:
         mode: lint
         repository-list: ${{ needs.setup.outputs.repository-list }}
         tool-list: ${{ needs.setup.outputs.tool-list }}
-        additional-planemo-options: --biocontainers -s tests,output,inputs,help,general,command,citations,tool_xsd,xml_order,tool_urls,shed_metadata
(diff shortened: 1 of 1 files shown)


francoismg:
@dsavchenko @volodymyrss

dsavchenko:
I tried this change [directly](https://github.com/esg-epfl-apc/tools-astro/pull/208/commits/68ee4ad8b7294ea81c979bb4e9ac15fe04ff1731), and it doesn't help. 
I think this check is activated by `--urls` flag [here](https://github.com/galaxyproject/planemo-ci-action/blob/fb7af7555194b70a93637a530cafd869e1f5e6c3/planemo_ci_actions.sh#L119), and it's not configurable

dsavchenko:
as this approach doesn't work, I close

PR #234, opened 2025-07-01
fits2bitmap output format selector

### LINKED ISSUE #232: make fits2tiff galaxy tool

dsavchenko:


tools/astropytools/fits2bitmap.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="astropy_fits2bitmap" name="astropy fits2bitmap" version="0.2.0+galaxy1" profile="21.05">
+<tool id="astropy_fits2bitmap" name="astropy fits2bitmap" version="0.3.0+galaxy0" profile="24.0">
```
Andrei-EPFL, line 1: I do not understand what galaxy0,galaxy1,galaxy2 mean, but I am not sure I need to understand if they make sense as they are. It seems that each file has a different value.

dsavchenko:
> thanks! Super nice! But I realized something: this fits2bitmap will compress the information from the FITS file. whereas a standard TIFF file could support the same precision as in the FITS file, i.e. float numbers.

Do you have a reference that it compresses an image? Just for me to understand

tools/astropytools/fits2bitmap.xml
dsavchenko: This is a part of a version string, additional to the semver. It's intended to be used to distinguish the versions of the tool with minor changes in the tool description xml that doesn't affect tool functionality. 
Here, I bump the version of every tool in a tool suite because otherwise linting fails. Not sure about the reasons, but this hack works.

dsavchenko:
@volodymyrss please review

volodymyrss (approved):

